PR #3, opened 2018-10-24
Allow setting topological target at build time.

Fryguy:


Fryguy:
I haven't tested the build yet

Fryguy:
I've tested the build and the resultant javascript looks correct.  While I can't get the actual deployment working yet for other reasons, this code should work as is.

@lgalis Please review.

PR #5, opened 2018-10-25
Remove API client code, Order APIs updated for Portfolio items, styling and path updates.

lgalis:
Update styling, remove extra Grid and GridItem elements.
Update paths and package.json to match insights updates.
Update relative paths for imports.
Remove the Auto-generated swagger API client code for the topological_inventory and service portal. (these were moved to separate repos.)
Update the Order ui code with the new API calls applying to a  portfolio item.

After:
screenshot from 2018-11-01 09-45-04


gmcculloug:
@lgalis Can you add a pre/post images for the updated styling.  Any thoughts on who could review?

lgalis:
There are no tests yet - will update the webpack loading that is currently causing the build to show as failing in a separate PR.

PR #6, opened 2018-11-02
Fix styling build errors

lgalis:
- test -


lgalis:
@gmcculloug - fixed styling errors in this PR - will fix the codeclimate ones in a following PR

PR #10, opened 2018-11-08
Changes to use the new insights-dev Jenkins

lgalis:
Changes to use the new insights-dev Jenkins

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ script:
 - npm run travis:verify
 after_script:
 - "./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT"
+after_success:
+curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
 deploy:
   provider: script
   script: bash .travis/deploy.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,14 +17,9 @@ script:
 after_script:
 - "./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT"
 after_success:
-curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -16,8 +16,15 @@ script:
 - npm run travis:verify
 after_script:
 - "./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT"
+after_success:
+curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
```
Fryguy: this should not be a separate line, and if it is, it can't be undented like this (invalid yaml)...might be better to do this

```suggestion
- curl -sSL https://raw.githubusercontent.com/RedHatInsights/insights-frontend-builder-common/master/src/bootstrap.sh | bash -s
```

.travis.yml
```diff
@@ -16,8 +16,15 @@ script:
...
 deploy:
   provider: script
   script: bash .travis/deploy.sh
   'on':
     branch: master
```
Fryguy: I think this has to change as well to not use deploy.sh.

.travis.yml
```diff
@@ -16,8 +16,15 @@ script:
...
+env:
+    global:
+    - REPO="git@github.com:RedHatInsights/insights-service-portal-build"
```
Fryguy: I don't like having this in the repo, but I'm assuming this is needed by the downloaded script?  I'd prefer to keep this as a travis-side encrypted env var, and not baked into the file.

Also (and this may not be for this PR) *most* of this information can be derived in a script like a I did here: https://github.com/ManageIQ/service_portal-ui/blob/master/.travis/release.sh#L3-L4 (note that DEPLOY_REPO is a build time env var, so I can change where to point to and not bake it into the .travis.yml)

@iphands ... can we chat about this?

Fryguy:
@lgalis Did you run the migrate.sh script to convert this repo?

gmcculloug:
@lgalis See question above: https://github.com/ManageIQ/service_portal-ui/pull/10#issuecomment-437137839

@Fryguy Is there anything other changes needed here?

lgalis:
@Fryguy, @gmcculloug - no, I did not use the script

PR #12, opened 2018-11-13
Add All Portfolios Page and Single Portfolio page

lgalis:
All Portfolios Page:

screenshot from 2018-11-20 10-15-10

Portfolio with actions toolbar:
screenshot from 2018-11-20 15-32-45


package.json
```diff
@@ -19,7 +19,6 @@
     "react-modal": "^3.5.1",
     "react-spinners": "^0.3.2",
     "react-transition-group": "^2.4.0",
-    "redux-logger": "^3.0.6",
```
hstastna: What is the reason to remove redux logger?

src/PresentationalComponents/Portfolio/portfoliocard.scss
```diff
@@ -0,0 +1,51 @@
+@import "../../App.scss";
+
+.pcard_style {
+  margin-top: 0px;
```
hstastna: What is the reason to put every option to the single line and define the values like this?
Maybe

margin: 0px;
padding: 0px;

could work, too. https://www.w3schools.com/Css/css_margin.asp

package.json
lgalis: it was a merge issue, it's back in

src/PresentationalComponents/Portfolio/PorfolioCard.js
```diff
@@ -0,0 +1,37 @@
...
+            <GridItem GridItem sm={ 6 } md={ 4 } lg={ 4 } xl={ 3 }>
+                <Card className="pcard_style">
+                    <CardHeader className="pcard_header">
+                        <ImageWithDefault src={ this.props.imageUrl || DefaultPortfolioImg } defaultSrc={ DefaultPortfolioImg } />
+                    </CardHeader>
+                    <CardBody className="pcard_body">
+                        <h4>{ this.props.name }</h4>
+                        { itemDetails(this.props, TO_DISPLAY) }
+                    </CardBody>
+                    <CardFooter/>
```
hstastna: Do we need  `<CardFooter/>` here?

src/Store/ActionTypes.js
```diff
@@ -19,7 +19,7 @@ export const FILTER_PLATFORM_ITEMS = 'FILTER_PLATFORM_ITEMS';
 export const FETCH_PORTFOLIOS = 'FETCH_PORTFOLIOS';
 export const FETCH_PORTFOLIO_ITEMS = 'FETCH_PORTFOLIO_ITEMS';
 export const FETCH_PORTFOLIO_ITEMS_WITH_PORTFOLIO = 'FETCH_PORTFOLIO_ITEMS_WITH_PORTFOLIO';
-export const FETCH_PORTFOLIO_ITEM = 'FETCH_PORTFOLIO_ITEM';
+export const FETCH_PORTFOLIO = 'FETCH_PORTFOLIO';
```
hstastna: It's better to divide action types into more files, not to have them in only one file like this. The same as we have actions in `AlertActions.js`,`OrderActions.js`, `PlatformActions.js`,... We don't have them all in something like `Actions.js` (and the same for reducers) :)

src/PresentationalComponents/Portfolio/PorfolioCard.js
lgalis: Not at the moment, but we'll need to add details, so I will leave it in.

src/Store/ActionTypes.js
lgalis: We only have a few at the moment,. Agree, as we have more actions, we should divide the file into store categories.

lgalis:
@hstatsna - I didnot notice the latest UX has teh actions switched - thanks - I will change that in the next PR with the rest of the toolbars.

lgalis:
Most of the codeclimate warnings are caused by the React's propTypes declarations being marked as duplicate. Codeclimate has issue #89 opened for this, with some changes made to allow to filter these out. We could look into configuring a filter for these.

src/Store/Actions/PortfolioActions.js
```diff
@@ -40,7 +40,12 @@ export const searchPortfolioItems = value => ({
...
 
-export const addPortfolioWithItem = (portfolioData, itemData) => ({
+export const addPortfolio = (portfolioData, items) => ({
     type: ActionTypes.ADD_PORTFOLIO,
-    payload: PortfolioHelper.addPortfolioWithItem(portfolioData, itemData)
+    payload: PortfolioHelper.addPortfolio(portfolioData, items)
+});
+
+export const updatePortfolio = (portfolioData) => ({
+    type: ActionTypes.UPDATE_PORTFOLIO,
```
hstastna: You are missing UPDATE_PORTFOLIO in ActionTypes.

src/Store/ActionTypes.js
hstastna: If you do such changes, it is also good to remove references also in _src/Store/Reducers/PortfolioStore.js_.

src/Store/Actions/PortfolioActions.js
lgalis: The API for this does not exist yet - I will add the Action type when the API is actually available.

hstastna: No worries, I already added the type in my actual open PR => an ugly warning disappeared and everything works as before.

PR #16, opened 2018-11-26
Update Navigation list for empty platform/portfolio list 

lgalis:
Update Navigation list for empty platform/portfolio list 
Remove the Edit icon in the navigation list as it is removed from the new UX design.


lgalis:
@gmcculloug  - please merge when you can - this allows us to use the UI when the topology service is not available

PR #18, opened 2018-12-05
Update path to use service-portal

lgalis:


Fryguy:
Going to merge as is without CI because CI is broken due to the npm event-stream thing.

PR #25, opened 2018-12-10
More fixes for deployment

Fryguy:
@gmcculloug Please review.

@lgalis This includes the change for the service_portal const.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,12 @@ node_modules
 # production
 dist
 
+# deployment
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -12,6 +12,12 @@ node_modules
 # production
 dist
 
+# deployment
+.travis/58231b16fdee45a03a4ee3cf94a9f2c3
```
gmcculloug: For future reference this file comes from https://github.com/RedHatInsights/insights-frontend-builder-common/blob/master/src/bootstrap.sh#L8

PR #33, opened 2019-01-02
Fixed menu min height to stretch from top to bottom.

Hyperkid123:
Stretched navigation min height from to bottom of the screen to match designs.


Hyperkid123:
Travis should be fixed by #34

PR #35, opened 2019-01-02
Fix navigation

Hyperkid123:
### Changes
- Highlight active link
- No longer reloading nav on each link change


Hyperkid123:
Travis should be fixed by #34

PR #39, opened 2019-01-04
Portfolio modals

Hyperkid123:
Replaced redux controlled modals for adding/editing portfolio.

### Changes
- unique route for portfolio items
  - cannot share the same route as portfolios
  - it will break any nested routing for portfolios (some other views, filtering etc...)
- merged modal for add/editing into one component
- modals are now controlled via react routing not redux

I also wanted to write test for this, but there are some issues with the client APIs (they are not transpiled to ES6).

I will fix that after i'm back from PTO.

This will probably need some clicking through the app to make sure i did not break anything.


Hyperkid123:
@miq-bot remove-label unmergeable

PR #41, opened 2019-01-04
Add back the addToOrder call removed by refactoring

lgalis:
Updated type for servicePlans.


lgalis:
@mkanoor - please review

PR #43, opened 2019-01-07
Sort orders by descending create date

lgalis:
Sort descending by create date


lgalis:
@gmcculloug - sort for orders  - this component will be redone based on the UX design - this is a placeholder

PR #45, opened 2019-01-07
Restore portfolio list refresh when adding a portfolio.

lgalis:
Restore the refresh portfolio list after add - this was removed during refactoring


lgalis:
@gmcculloug - please review

PR #47, opened 2019-01-10
Api clients package

Hyperkid123:
I've created npm packages for Sources and for SSP API clients. We can now share our API with other teams.

### Important
The sources api is a bit broken right now. It will not work if you run your app from `prod.foo...` domain. You have to run it from `https://ci.foo.redhat.com:1337/insights/platform/service-portal/...` in order to work properly (it seems that its not deployed for some reason).

I've been talking about this with MartinP and he is comunicating this issue to the sources API devs. It should be resolved in near future.

### Packages:
- service portal: https://www.npmjs.com/package/@manageiq/service-portal-api
- topology inventory: https://www.npmjs.com/package/@manageiq/topological_inventory

There were some changes required to make it work.

### Changes
- upgrade to babel 7
- replace babel.config.js with .babelrc
  - this should not have work in first place
  - babel.config.js is not valid config file for babel and it was not included anywhere else
- add babel bridge resolution
  - makes jest (tests) work
- use npm packages form API clients
- configure basePath to SSP in ui client


src/Utilities/Constants.js
```diff
@@ -1,2 +1,8 @@
 export const DEFAULT_PAGE_SIZE = 50;
+export const SOURCES_API_BASE = '/r/insights/platform/topological-inventory/v0.0';
```
lgalis: const topo_apiHost = process.env.TOPO_API_HOST || '';
const topo_apiPort = process.env.TOPO_API_PORT || '443';
const topo_basePath = process.env.TOPO_BASE_PATH;
const topo_apiVersion =  process.env.TOPO_API_VERSION || '/api/v0.0';
export const SOURCES_API_BASE = `https://'${topo_apiHost}:${topo_apiPort}${topo_basePath}${topo_apiVersion.replace(/\/+$/, '')}`;

lgalis:
Rather then a separate PR - @Hyperkid123 ,  it would be better if you make the 2 changes I mention in your PR here.

Hyperkid123:
@lgalis thanks for the feedback. I've applied you suggestions.

Unfortunatelly there seems to be some issue with the topological API so i've added wip label until a can figure out what it is.

src/Utilities/Constants.js
Hyperkid123: > const topo_apiHost = process.env.TOPO_API_HOST || '';
> const topo_apiPort = process.env.TOPO_API_PORT || '443';
> const topo_basePath = process.env.TOPO_BASE_PATH;
> const topo_apiVersion = process.env.TOPO_API_VERSION || '/api/v0.0';
> export const SOURCES_API_BASE = `https://'${topo_apiHost}:${topo_apiPort}${topo_basePath}${topo_apiVersion.replace(/\/+$/, '')}`;

Ok after small discussion with @karelhala we ought not to use full URI path for our APIs. host port etc. is retrieved from origin evrionment because API can run on multiple servers. 
There are 3 envirnoments where the APIs live: CI, QA, Prod. If we hard code base path from host we will break the UI on other server.

Also the base path you've suggested is missing some fragments of the URL

The reason why is it working for the service portal API is that we are running it locally.

> Unfortunatelly there seems to be some issue with the topological API so i've added wip label until i figure out what it is.

After i've removed your suggestion it is working OK.

Hyperkid123:
Also the service portal api seems to have wrong configuration. The api URL should not be something like  `/api/v0.0` but it ought to be `/r/insights/platform/service-portal/v0.0` to match other API clients even for local use. That way we can use insights proxy to api requests.

Hyperkid123:
You may also notice slower performance for the sources API. But that is actually closer to real world performance. To offset this we can use pattern called optimistic UI and optimize data retrieval. But that is also something for the future. Not important right now

Fryguy:
> Also the service portal api seems to have wrong configuration. The api URL should not be something like /api/v0.0 but it ought to be /r/insights/platform/service-portal/v0.0 to match other API clients even for local use. That way we can use insights proxy to api requests.
> 
> Redirecting from proxy to locally running API can be done via SPANDX_CONFIG. But that is something that should be done in other PRs and will probably need some changes to the API repo.

@gmcculloug @Hyperkid123 The topological api is not wrong.  It is driven by an env var BASE_PATH which the insights platform will give us as part of the deployment (currently configured as part of the OCD deployment template).  This is so that the platform can be configured and deployed without code changes.  In the absence of that env var, the service defaults /api/ as the BASE_PATH.  Note that the intention for many of the services is for eventual reuse in other applications, so hardcoding anything about the base path is incorrect.  Unfortunately the way swagger works we are *forced* to hardcode the base path there, so that client generation will work.  There are possible ways around that, but we;ve chosen to hardcode that one place for now.

PR #52, opened 2019-01-14
Platforms and portfolios data loading

Hyperkid123:
Currently we have overlapping data loading for platforms and portfolios. These are loaded from multiple components (Navigation, Platforms/Portfolio). This means that we have two same API requests running at the same time and overwriting each other.

screenshot from 2019-01-14 12-49-47

I've added simple check that will allow these requests only if there isn't any other in progress (using the isLoading flag).

Note: This is not 100% solution, because we are lazy loading our components and there is no system in place that invalidates data in redux. Sometimes it will happen that the requests will trigger twice, because one request will finish before another is triggered (when you run API locally for instance).

This can be resolved sometime in future when we cal invalidate our data.

### Changes
- order of middlewares
  - thunk must be place before promise middleware
- using new action to load platform/portfolios
  - checks if another request is in progress


Hyperkid123:
@miq-bot remove-label unmergeable

PR #54, opened 2019-01-14
Handle the case when service-portal is considered an unknown app identifier in chrome

lgalis:


lgalis:
@gmcculloug  = please review

PR #56, opened 2019-01-15
Remove package-lock.json from gitignore

lgalis:
... it was added during refactoring


Hyperkid123:
@lgalis i would probably remove yarn.lock from ignore too if we want to preserve dependencies versions. Yarn and npm are using different lock files.

PR #67, opened 2019-01-21
[WIP] Update nav to match mocks.

Hyperkid123:
- Added the indentation to sub items in navigation
- Added products count to portfolio nav

### NOTE: products count is just a placeholder until API is available. For now its always 0.

screenshot-ci foo redhat com-1337-2019 01 21-16-51-46

screenshot-ci foo redhat com-1337-2019 01 21-16-54-02


Hyperkid123:
> This looks good, however, I think we should only merge this once the backed supports retrieving the count information, Having a 0 is misleading for the user.

Agreed renaming this to WIP

PR #80, opened 2019-01-25
Cleanup unused env vars and rename for clarity

Fryguy:
Additionally, use the BASE_PATH env var as part of the API paths

@martinpovolny @Hyperkid123 Please review.  Let's also discuss before merge because I need to change the travis environment to have BASE_PATH set properly (and less importantly, so I can remove the rest of the env vars defined there).

### DIFF
--- a/config/base.webpack.plugins.js
+++ b/config/base.webpack.plugins.js
@@ -87,14 +87,7 @@ plugins.push(CopyFilesWebpackPlugin);
  * Makes build-time env vars available to the client-side as constants
  */
 const envPlugin = new webpack.DefinePlugin({
-    'process.env.API_HOST': JSON.stringify(process.env.API_HOST),
-    'process.env.API_PORT': JSON.stringify(process.env.API_PORT),
-    'process.env.BASE_PATH': JSON.stringify(process.env.BASE_PATH),
-    'process.env.API_VERSION': JSON.stringify(process.env.API_VERSION),
-    'process.env.TOPO_API_HOST': JSON.stringify(process.env.TOPO_API_HOST),
-    'process.env.TOPO_API_PORT': JSON.stringify(process.env.TOPO_API_PORT),
-    'process.env.TOPO_BASE_PATH': JSON.stringify(process.env.TOPO_BASE_PATH),
-    'process.env.TOPO_API_VERSION': JSON.stringify(process.env.TOPO_API_VERSION)
+    'process.env.BASE_PATH': JSON.stringify(process.env.BASE_PATH || '/r/insights/platform')
 });
 plugins.push(envPlugin);
 
--- a/src/Helpers/Shared/userLogin.js
+++ b/src/Helpers/Shared/userLogin.js
@@ -1,11 +1,11 @@
 import { AdminsApi, ApiClient as ServicePortalApiClient } from '@manageiq/service-portal-api';
 import { DefaultApi, ApiClient as TopologicalInventoryApiClient } from '@manageiq/topological_inventory';
-import { SOURCES_API_BASE, SERVICE_PORTAL_API_BASE } from '../../Utilities/Constants';
+import { TOPOLOGICAL_INVENTORY_API_BASE, SERVICE_PORTAL_API_BASE } from '../../Utilities/Constants';
 
 const adminApi = new AdminsApi();
 
 const defaultClient = TopologicalInventoryApiClient.instance;
-defaultClient.basePath = SOURCES_API_BASE;
+defaultClient.basePath = TOPOLOGICAL_INVENTORY_API_BASE;
 
 const sspDefaultClient = ServicePortalApiClient.instance;
 sspDefaultClient.basePath = SERVICE_PORTAL_API_BASE;
--- a/src/Utilities/Constants.js
+++ b/src/Utilities/Constants.js
@@ -1,3 +1,2 @@
-
-export const SOURCES_API_BASE = `/r/insights/platform/topological-inventory/v0.0`;
-export const SERVICE_PORTAL_API_BASE = `/r/insights/platform/service-portal/v0.0`;
+export const SERVICE_PORTAL_API_BASE = `${process.env.BASE_PATH}/service-portal/v0.0`;
+export const TOPOLOGICAL_INVENTORY_API_BASE = `${process.env.BASE_PATH}/topological-inventory/v0.0`;
--- a/src/test/redux/actions/portfolio-actions.test.js
+++ b/src/test/redux/actions/portfolio-actions.test.js
@@ -18,6 +18,9 @@ import {
   updatePortfolio,
   removePortfolio
 } from '../../../redux/Actions/PortfolioActions';
+import {
+  SERVICE_PORTAL_API_BASE
+} from '../../../Utilities/Constants';
 
 describe('Portfolio actions', () => {
   const middlewares = [ thunk, promiseMiddleware(), notificationsMiddleware() ];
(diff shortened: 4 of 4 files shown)


src/Helpers/Shared/userLogin.js
```diff
@@ -1,11 +1,11 @@
 import { AdminsApi, ApiClient as ServicePortalApiClient } from '@manageiq/service-portal-api';
 import { DefaultApi, ApiClient as TopologicalInventoryApiClient } from '@manageiq/topological_inventory';
-import { SOURCES_API_BASE, SERVICE_PORTAL_API_BASE } from '../../Utilities/Constants';
+import { TOPOLOGICAL_INVENTORY_API_BASE, SERVICE_PORTAL_API_BASE } from '../../Utilities/Constants';
 
 const adminApi = new AdminsApi();
 
 const defaultClient = TopologicalInventoryApiClient.instance;
-defaultClient.basePath = SOURCES_API_BASE;
+defaultClient.basePath = TOPOLOGICAL_INVENTORY_API_BASE;
```
Fryguy: One weird thing here is that this defaultClient is not used in the getTopologicalUserApi method below, which is confusing.  Instead the method uses DefaultApi from '@manageiq/topological_inventory'. Are we setting a base path in something that isn't used?  Same thing for the sspDefaultClient

src/Helpers/Shared/userLogin.js
```diff
@@ -1,11 +1,11 @@
 import { AdminsApi, ApiClient as ServicePortalApiClient } from '@manageiq/service-portal-api';
 import { DefaultApi, ApiClient as TopologicalInventoryApiClient } from '@manageiq/topological_inventory';
```
Fryguy: Minor, but the package names are inconsistent.   I know we haven't moved over the service-portal-api client to the ManageIQ repo, but I expected the client package names to be consistent.

config/base.webpack.plugins.js
```diff
@@ -87,14 +87,7 @@ plugins.push(CopyFilesWebpackPlugin);
...
 const envPlugin = new webpack.DefinePlugin({
-    'process.env.API_HOST': JSON.stringify(process.env.API_HOST),
-    'process.env.API_PORT': JSON.stringify(process.env.API_PORT),
-    'process.env.BASE_PATH': JSON.stringify(process.env.BASE_PATH),
-    'process.env.API_VERSION': JSON.stringify(process.env.API_VERSION),
-    'process.env.TOPO_API_HOST': JSON.stringify(process.env.TOPO_API_HOST),
-    'process.env.TOPO_API_PORT': JSON.stringify(process.env.TOPO_API_PORT),
-    'process.env.TOPO_BASE_PATH': JSON.stringify(process.env.TOPO_BASE_PATH),
-    'process.env.TOPO_API_VERSION': JSON.stringify(process.env.TOPO_API_VERSION)
+    'process.env.BASE_PATH': JSON.stringify(process.env.BASE_PATH) || '/r/insights/platform'
```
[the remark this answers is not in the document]
lgalis: i think there should be no default here, and instead have a default, if needed, where these variables are used.

src/Helpers/Shared/userLogin.js
Hyperkid123: @Fryguy event though the ApiClients are not used directly, they are called byt adminsApi/usersApi. It is a singleton that should be used to configure your apps. That way you can change the basePaths and you don't have to hardcode them inside the clients.

src/Utilities/Constants.js
```diff
@@ -1,3 +1,2 @@
-
-export const SOURCES_API_BASE = `/r/insights/platform/topological-inventory/v0.0`;
-export const SERVICE_PORTAL_API_BASE = `/r/insights/platform/service-portal/v0.0`;
+export const SERVICE_PORTAL_API_BASE = process.env.BASE_PATH + '/service-portal/v0.0';
+export const TOPOLOGICAL_INVENTORY_API_BASE = process.env.BASE_PATH + '/topological-inventory/v0.0';
```
Hyperkid123: i think the '/r' path is required. It marks the path as an API.

src/Helpers/Shared/userLogin.js
Hyperkid123: @Fryguy i agree

src/Utilities/Constants.js
lgalis: Should we have environment variables for /service-portal/v0.0 and  /topological-inventory/v0.0?

Fryguy:
@Hyperkid123 were you able to get back to this?

src/Utilities/Constants.js
Fryguy: Maybe?  It's probably smart to have that and default it when not set.

config/base.webpack.plugins.js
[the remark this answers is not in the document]
lgalis: I would take the default out from here and move it to where the BASE_PATH is used.

src/Utilities/Constants.js
lgalis: This should be:

onst base_path = process.env.BASE_PATH || '/r/insights/platform';

export const SERVICE_PORTAL_API_BASE = `${base_path}/service-portal/v0.0`;
export const TOPOLOGICAL_INVENTORY_API_BASE = `${base_path}/topological-inventory/v0.0`;

Hyperkid123:
@Fryguy OK se the issue is was very simple. The default basePath must be stringyfied otherwise it is not a string but a regular expression. Can be fixed at `base.webpack.plugins` with this small patch:

```diff
diff --git a/config/base.webpack.plugins.js b/config/base.webpack.plugins.js
index 3c822a1..91cc7f9 100644
--- a/config/base.webpack.plugins.js
+++ b/config/base.webpack.plugins.js
@@ -87,7 +87,7 @@ plugins.push(CopyFilesWebpackPlugin);
  * Makes build-time env vars available to the client-side as constants
  */
 const envPlugin = new webpack.DefinePlugin({
-    'process.env.BASE_PATH': JSON.stringify(process.env.BASE_PATH) || '/r/insights/platform'
+    'process.env.BASE_PATH': JSON.stringify(process.env.BASE_PATH) || JSON.stringify('/r/insights/platform')
 });
 plugins.push(envPlugin);
```

Hyperkid123:
Apart from the small issue it looks good. After the patch API still works. I've got just a small nitpicks.

src/Utilities/Constants.js
```diff
@@ -1,3 +1,2 @@
-
-export const SOURCES_API_BASE = `/r/insights/platform/topological-inventory/v0.0`;
-export const SERVICE_PORTAL_API_BASE = `/r/insights/platform/service-portal/v0.0`;
+export const SERVICE_PORTAL_API_BASE = process.env.BASE_PATH + '/service-portal/v0.0';
```
Hyperkid123: Can you please use template strings. Its just about code styling.
```javascript
export const SERVICE_PORTAL_API_BASE = `${process.env.BASE_PATH}/service-portal/v0.0`;
```

src/Utilities/Constants.js
Fryguy: I meant to put the default inside the parens...just a bug.

src/Utilities/Constants.js
Fryguy: I really don't understand template strings.  Can you explain why they are preferable over normal strings or give me a link with a good explanation?

Fryguy:
> The default basePath must be stringyfied otherwise it is not a string but a regular expression. Can be fixed at `base.webpack.plugins` with this small patch:

Ah yeah, I meant to put the || inside the parens... see https://github.com/ManageIQ/service_portal-ui/pull/80#discussion_r253582126

src/Utilities/Constants.js
Hyperkid123: @Fryguy for this case its just a different syntax. But it has some neat features especially for string interpolation. Also the `+` is ambiguous and in some cases may have bad side effects (if you don't have 100% control over the variable content). That won't happen in this case obviously but we are using it across the code and i would like to keep it consistent.

There is a lot of articles across the internet. But this one is good imo: https://developers.google.com/web/updates/2015/01/ES6-Template-Strings

PR #81, opened 2019-02-04
Added new pf classes to app template.

Hyperkid123:
Insights page layout will be changed a bit in a future. This PR prepares PF classes to transition once the changes in Chrome are ready. Has no effect right now.


karelhala:
:shipit:

PR #82, opened 2019-02-05
Added selection of destination project for order modal

Hyperkid123:
### Need info
- Are these supposed to show for every portfolio or only for some kind. If so, how can i tell when to make this request?
- Do we know where the select field should be? On the top/bottom somewhere else?
- Thanks to the nested structure of JSON schema, output data format of the for for the destination will be nested. What is the expected data format? Is it just a random object? Right now the generated docs does not specify how the data should look like.
```javascript
{
  name: 'Foo',
  description: 'Bar',
  providerControlParameters: {
    namespace: 'default'
  }
}
``` 

### Issues
- New API clients returns instances of classes. That will hurt us in a long run. Right now it enforces using classes in only tests. This would be fine if we are creating these objects in UI but everything is done via API and will add additional overhead.
- pf4 changes styling for multiple components. Modal and select are broken once again
- PF4 select component does not recognize empty value and will always select first option, but that will not update form state, that is why the select is in error state even when `default` (first) option is selected. And because we are using mozilla JSON schema which uses the enum with just values (not combo label, value) we cannot add some `Please Choose` option with empty(undefined) value.
- The helper function from API client for retrieving the schema is always empty object. It returns a readable stream that is not converted to JSON inside the API client. I had to use custom `fetch` request to return correct data. Making the API to return only plain JSON objects would fix this issue.
- Options for destination contain duplicate values. Created a SET from the array, that fixed the problem. But i think the API should not return duplicate values.

dynamic-form

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -861,17 +861,17 @@
       }
     },
     "@data-driven-forms/pf4-component-mapper": {
-      "version": "0.3.2",
-      "resolved": "https://registry.npmjs.org/@data-driven-forms/pf4-component-mapper/-/pf4-component-mapper-0.3.2.tgz",
-      "integrity": "sha512-NEVbEELIl6wKfBp/h0E+hHfUHqrmeT8G+dqDb3mOybN1nzl8q1cgQaBWmRfhRJPoMIq7SNtH9y6SJCn3wr4SRw==",
+      "version": "0.4.1",
+      "resolved": "https://registry.npmjs.org/@data-driven-forms/pf4-component-mapper/-/pf4-component-mapper-0.4.1.tgz",
+      "integrity": "sha512-8+CS9Vz9PLBqc5NHlMgxW66UNBh51pURSIQvmtvIW4Tgsf95SejnkBKH5iuVNymSnef5ilxNHnGXxYWpCfkVKw==",
       "requires": {
         "@patternfly/patternfly-next": "^1.0.87"
       }
     },
     "@data-driven-forms/react-form-renderer": {
-      "version": "1.5.5",
-      "resolved": "https://registry.npmjs.org/@data-driven-forms/react-form-renderer/-/react-form-renderer-1.5.5.tgz",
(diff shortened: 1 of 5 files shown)


gmcculloug:
> * Are these supposed to show for every portfolio or only for some kind. If so, how can i tell when to make this request?

It should show when ordering any portfolio item.  Since we only support OCP today it will always be populated.  We will need to evaluate future Platforms to determine if it is needed for each item and how to control when it is not needed.

> * Do we know where the select field should be? On the top/bottom somewhere else?

Top of the form would match current OCP design.

> * Thanks to the nested structure of JSON schema, output data format of the for for the destination will be nested. What is the expected data format? Is it just a random object? Right now the generated docs does not specify how the data should look like.

@mkanoor Thoughts here?

Hyperkid123:
> * Thanks to the nested structure of JSON schema, output data format of the for for the destination will be nested. What is the expected data format? Is it just a random object? Right now the generated docs does not specify how the data should look like.

So what i mean by this. Is there any data format requirement? From what i can tell the data nesting is how it should look like but i don't know what attributes should be there. I assume only the stuff that is inside the form but you never know.

gmcculloug:
For OCP items, which is the only type supported at the moment, there should always be a project or list of projects available.  It is a required field so you should not be able to order without filling it in.

A default value would be nice, but I'm not sure we can identify that today unless there is an actual "default" project coming back from OCP.  I do not know if that will always be the case on every OCP environment.

As for the issue @lgalis ran into above, it appears that we were once collecting the `source_id` to populate the `service_offering_source_ref` for the PortfolioItem but now we collect the `source_ref` which is the incorrect value.

```yaml
{
  id: "88747",
  name: "postgresql-persistent",
  display_name: "PostgreSQL",
  source_ref: "7ce0f818-fdee-11e8-860c-06945c5af756",
  source_id: "4",
  service_offering_icon_id: "4",
  ...
}
```

Forcing the proper value of 4 into the database caused projects to display as expected.

We'll create a PR to address this on the service portal api.  cc @syncrou

PR #83, opened 2019-02-05
Tests for platform actions.

Hyperkid123:
Added set of tests for redux platform action creators.


src/test/redux/actions/platform-actions.test.js
```diff
@@ -0,0 +1,270 @@
...
+    mockStore = configureStore(middlewares);
+  });
+
+  it('should dispatch correct actions after fetching platforms', () => {
+    const store = mockStore({
+      platformReducer: {
+        isPlatformDataLoading: false
+      }
+    });
+    apiClientMock.get('/r/insights/platform/topological-inventory/v0.0/sources', mockOnce({
```
lgalis: we should use the TOPOLOGICAL_INVENTORY_API_BASE here

src/test/redux/actions/platform-actions.test.js
```diff
@@ -0,0 +1,270 @@
...
+    });
+  });
+
+  it('should dispatch correct actions after fetching multiple platforms items fails', () => {
+    const store = mockStore({
+      platformReducer: {
+        isPlatformDataLoading: false
+      }
+    });
+    apiClientMock.get('/r/insights/platform/topological-inventory/v0.0/sources/1/service_offerings', mockOnce({
```
lgalis: use TOPOLOGICAL_INVENTORY_API_BASE  here - and the rest of the places in this file where the base path is used

Hyperkid123: Ah nice catch!

PR #84, opened 2019-02-05
Use PF4 data list to show orders.

Hyperkid123:
Replaces the Insights table with PF4 data list to match design mocks. Is still mostly a placeholder but can be easily updated once we have necessary data.

### Before
screenshot from 2019-02-05 14-47-02

### After
order-data-list

Edit: I'm aware of the duplicate code. Will be resolved once we can differentiate between all orders and my orders.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1051,9 +1051,9 @@
       "dev": true
     },
     "@manageiq/service-portal-api": {
-      "version": "1.1.1",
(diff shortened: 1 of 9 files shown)


Hyperkid123:
Travis is fixed by https://github.com/ManageIQ/service_portal-ui/pull/82

PR #85, opened 2019-02-06
Update api 0.1.0

Hyperkid123:
Updated service catalog and topology API version to schema version 0.1.0

### Changes
- locked js API client version to patch releases only (Bugfixes etc.)
- Pagination data is not stored yet. Will come later when we actually start paginate. It will required some redux store changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1051,9 +1051,9 @@
       "dev": true
     },
     "@manageiq/service-portal-api": {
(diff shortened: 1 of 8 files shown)


Hyperkid123:
waiting for https://github.com/ManageIQ/topological_inventory-api-jsclient/pull/4 to update the topological API calls.

PR #89, opened 2019-02-11
Fixed unauthorized error

Hyperkid123:
Fixes issue with unauthorized error after login. 

### Issue
Chrome sends authorization TOKEN to ui before it is sent to APIs. So the UI loads but the user is not authorized in APIs so the requests will throw an error.

### Changes
Adds a workaround. UI waits on auth token from chrome (the same that is sent to API) and shows placeholder until auth token is received. This should be solved in chrome , but because they don't own the auth service they cannot do this right now. Maybe they will come up with something in a future.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,14 +12,17 @@ import { MIN_SCREEN_HEIGHT } from './constants/ui-constants';
 import '@red-hat-insights/insights-frontend-components/components/Notifications.css';
 
 import 'whatwg-fetch';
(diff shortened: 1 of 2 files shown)


gmcculloug:
@Hyperkid123 Is this how other apps on the platform resolved the issue?

lgalis:
@gmcculloug - yes, this is the recommended way -for now, wait until the token is received.

PR #96, opened 2019-02-13
[WIP] Initial pagination for platforms.

Hyperkid123:
Adds initial and very simple pagination for platform items.

### TODO
- fix prev/next requests
  - requires fix in topology API to send correct endpoints
  - offset is not computed from strings
- enable archived at as a attribute for `listSourceServiceOfferings`
  - more info: https://github.com/ManageIQ/topological_inventory-api-jsclient/blob/master/docs/DefaultApi.md#example-37
- screen designs
- filtering? Cannot combine server side pagination with client side filtering. Can be done on if all entities are given

### DIFF
--- a/src/Helpers/Platform/PlatformHelper.js
+++ b/src/Helpers/Platform/PlatformHelper.js
@@ -13,9 +13,9 @@ export function getPlatform(platformId) {
 
 export function getPlatformItems(apiProps, page) {
   let apiPromise = null;
-  console.log('apiProps: ', apiProps, page)
   if (apiProps) {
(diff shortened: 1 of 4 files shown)


gmcculloug:
I was thinking about the pagination/filtering issue as well.  It is definitely an issue but we might have to live with it for a little while until the apis have implemented filtering.

The pagination links in the json response should be usable now as they were updated to partial urls.

Hyperkid123:
> The pagination links in the json response should be usable now as they were updated to partial urls.

Yes i've just updated the links and they are working well.

> until the apis have implemented filtering.

I asked about the filter and so far you can "filter" only by entering full name (or other attribute) of some entity. For exmaple if i want entity with name "Foo" i can get that but if i want all entities with letter "f" somewhere in name i can't get that. So hopefully they will add substring search soon.

PR #97, opened 2019-02-14
Custom select

Hyperkid123:
Fixes issue with PF4 select not handling empty input value.

### Changes
- upgrade insights fronted components. Fixes transparent background
- Adds empty option to select if no value is selected
  - empty option with value `undefined` will not let user submit the form until he chooses something
  - why not add initial value? We cannot hard code initial value because we don't know what fields and values will be send from API. Also API response does not include default value.

select

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -81,7 +81,11 @@ rules:
     - warn
     - functions: false
   no-undef: 2
-  no-unused-vars: 2
+  no-unused-vars:
+    - error
+    - vars: all
+      args: after-used
+      ignoreRestSiblings: true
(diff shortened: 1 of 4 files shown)


gmcculloug:
@Hyperkid123 This update is a closer match to the OCP catalog which initially has "Select or create project" as the initial item.

Currently we only support selecting an existing project.

OCP example:
image

Hyperkid123:
@gmcculloug ideally i would like to just use placeholder property as is on the image you posted, but that is not supported yet so there is that empty option. Hopefully we will be able to get rid of it soon.

PR #104, opened 2019-02-19
Fixed order modal opening after removing portfolio item.

Hyperkid123:
### Issue
After removal of portfolio items modal for ordering shows up.

### Bug cause
- PF4 checkbox does not prevent bubbling of mouse events.
- Opening order modal and selecting items for removal is handled by same component and if you selected item for removal, the event bubbled down (should have been prevented by checkbox component) to the card which triggered opening of the modal.
- Also the modal opening was enabled even if the card was select able so i've also removed that option.

### Changes
- Added route for orders
- Order modal is now controlled via browser history instead of redux
- above change means that `MainModal` component is no longer used and was deleted.
- deleted all code used by `MainModal` component.


martinpovolny:
@Hyperkid123 : can you, please, file a bug against Patternfly on this?

Hyperkid123:
@martinpovolny Yes

martinpovolny:
The CC comments seem pedantic.

PR #105, opened 2019-02-20
Sort orders from newest to oldest.

Hyperkid123:
Orders are now sorted by date from newest do oldest directly in UI.

### DIFF
--- a/src/SmartComponents/Order/Orders.js
+++ b/src/SmartComponents/Order/Orders.js
@@ -26,7 +26,7 @@ class Orders extends Component {
 
   componentDidMount() {
     this.props.fetchOrders()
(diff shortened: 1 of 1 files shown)


src/SmartComponents/Order/Orders.js
```diff
@@ -26,7 +26,7 @@ class Orders extends Component {
 
   componentDidMount() {
     this.props.fetchOrders()
-    .then(() => this.setState({ orders: this.props.orderList.items }));
+    .then(() => this.setState({ orders: this.props.orderList.items.sort((a, b) => new Date(b.created_at) - new Date(a.created_at)) }));
```
gmcculloug: Can we use `id` (converted to an integer) for sorting?  Seems it would be less overhead than converting the date for each object.

PR #107, opened 2019-02-26
Product detail page.

Hyperkid123:
Adds detail page form portfolio items. Show all currently available data.

screenshot-ci foo redhat com-1337-2019 03 08-11-55-52

Did not include test just yet because most of this is probably going to change.

### DIFF
--- a/src/Helpers/Portfolio/PortfolioHelper.js
+++ b/src/Helpers/Portfolio/PortfolioHelper.js
@@ -15,8 +15,8 @@ export function listPortfolioItems() {
   return userApi.listPortfolioItems();
 }
 
-export function getPortfolioItem(portfolioId, portfolioItemId) {
-  return userApi.fetchPortfolioItemFromPortfolio(portfolioId, portfolioItemId);
+export function getPortfolioItem(portfolioItemId) {
+  return fetch(`${SERVICE_PORTAL_API_BASE}/portfolio_items/${portfolioItemId}`).then(data => data.json());
 }
 
 export function getPortfolio(portfolioId) {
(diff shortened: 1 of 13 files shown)


gmcculloug:
@Hyperkid123 Here is the story that documents the additional fields. https://projects.engineering.redhat.com/browse/SSP-128 (See attached images)

src/Helpers/Portfolio/PortfolioHelper.js
```diff
@@ -15,16 +15,16 @@ export function listPortfolioItems() {
   return userApi.listPortfolioItems();
 }
 
-export function getPortfolioItem(portfolioId, portfolioItemId) {
-  return userApi.fetchPortfolioItemFromPortfolio(portfolioId, portfolioItemId);
+export function getPortfolioItem(portfolioItemId) {
+  return fetch(`${SERVICE_PORTAL_API_BASE}/portfolio_items/${portfolioItemId}`).then(data => data.json());
```
lgalis: Are we switching to fetch here because of any issues with the openapi client?

Hyperkid123: No issue the client does not pass every attribute (Same as with the pagination). This is temporary until that is resolved.

Hyperkid123:
> We should double check with Susan on what the landing page should be post-order now that there is a Product detail page, and not a modal - I think it should be back to the Product Detail page.

I will definitely do that

src/Helpers/Portfolio/PortfolioHelper.js
lgalis: Sounds good, thx

gmcculloug:
> should we use 'PortfolioItemDetails' here rather than PortfolioProduct?

I would prefer that we avoid "marketing names" as these can change and stick to names that match the api endpoints.  "Product" is just the display name for "PortfolioItem"

Hyperkid123:
@lgalis ordering no longer redirects to orders page just closes the modal.

PR #110, opened 2019-03-01
Update toolbars

Hyperkid123:
Merge only after

Includes https://github.com/ManageIQ/service_portal-ui/pull/109

### TO DO
- [x] clean up commits after #109 was merged
- [x] clean up code
- [x] tests

### Fixes
- fixed delete link at portfolio card
- bugs cause by updating JS dependencies (caused over 2k changes in lock file)
  - fixed busted lodash plugin (added paths plugin and remove duplicate dependency)
  - fixed React components lazy loading (added node dynamic import plugin)

### Changes
- use one component for toolbars
- added generated breadcrumbs to toolbars

screenshot-ci foo redhat com-1337-2019 03 01-11-55-17


lgalis:
Not part of this PR, but the font for the product names seems too big, so some of the names are truncated - as can be seen above.

The remove dialog has truncated margins:

screenshot from 2019-03-04 12-25-01

Hyperkid123:
> The toolbar buttons are no longer right aligned as they are in the UX doc:

Yes i am aware of that. Some of the mocks have it on the left some on the right i will have to get more info on this.

Hyperkid123:
> Not part of this PR, but the font for the product names seems too big, so some of the names are truncated - as can be seen above.

PF4 changed modals again 😞 i can hard code it but i feel like this will be changing every day.

Hyperkid123:
@lgalis about the buttons. Just talked to Susan and from now on they will be aligned to the left.

lgalis:
@Hyperkid123 - left alignment sounds good - thanks for checking with UX.

PR #112, opened 2019-03-07
Platform pagination

Hyperkid123:
Added pagination for platform items. Its complete but just waiting for the API fix: https://github.com/ManageIQ/service_portal-api/pull/135 so i can make sure in the UI it does break anything even though we have covered it by tests.

### Changes
- store meta information about platform items collection
- use non PF4 paging component (will be swapped once PF4 has working implementation)
- updated card placeholder for loading to match gallery layout.
- resolving dependency issues again

platform-pagination

### DIFF
--- a/src/Helpers/Platform/PlatformHelper.js
+++ b/src/Helpers/Platform/PlatformHelper.js
@@ -11,11 +11,13 @@ export function getPlatform(platformId) {
   return api.showSource(platformId);
 }
 
-export function getPlatformItems(apiProps) {
+export function getPlatformItems(apiProps, options) {
   let apiPromise = null;
 
   if (apiProps) {
-    apiPromise = fetch(`${TOPOLOGICAL_INVENTORY_API_BASE}/sources/${apiProps}/service_offerings?archived_at=`).then(data => data.json());
+    apiPromise = fetch(`${TOPOLOGICAL_INVENTORY_API_BASE}/sources/${apiProps}/service_offerings?archived_at=${options
+      ? `&limit=${options.limit}&offset=${options.offset}`
+      : ''}`).then(data =>  data.json());
   } else {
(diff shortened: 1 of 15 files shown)


src/Helpers/Platform/PlatformHelper.js
```diff
@@ -11,11 +11,13 @@ export function getPlatform(platformId) {
   return api.showSource(platformId);
 }
 
-export function getPlatformItems(apiProps) {
+export function getPlatformItems(apiProps, options) {
   let apiPromise = null;
 
   if (apiProps) {
-    apiPromise = fetch(`${TOPOLOGICAL_INVENTORY_API_BASE}/sources/${apiProps}/service_offerings?archived_at=`).then(data => data.json());
+    apiPromise = fetch(`${TOPOLOGICAL_INVENTORY_API_BASE}/sources/${apiProps}/service_offerings?archived_at=${options
```
mkanoor: @Hyperkid123 Is this making a direct URL call instead of using the API?

lgalis: Wanted to check on the reason why we are using the fetch and not the generated api client

gmcculloug: @mkanoor Yes, it has been discussed and the api does not support filtering based on `archived_at` at the moment so we had to take this approach for now.

Hyperkid123: The client removes some attributes we need to get all relevant data (they are visible in the actual network response but the client removes them before it sends the response to the app). Until that is polished we need to sometimes use direct calls.

Hyperkid123: @mkanoor if i use the client method i cannot add anything but the pagination options to the request. I expect this to be resolved by the API some time in the future.

PR #113, opened 2019-03-08
Rename to catalog

Fryguy:
This will depend on 

- The build repo being renamed
- ~~The swagger clients rebuilt.~~

@gmcculloug


Fryguy:
I undid some of the package changes so that this PR doesn't have to be dependent on the client renames.

PR #114, opened 2019-03-11
Deploy to beta as well as stable

Fryguy:
@gmcculloug Please review.  Not sure if this is the specific set we want to push to, but it's a start on the beta line.


Fryguy:
@gmcculloug Only way to really test this is to merge and see what happens.

PR #115, opened 2019-03-12
Add products pagination

Hyperkid123:
### Changes
- now can select products only from one source at a time
- adds pagination when selecting products
- refactoring AddProductsToPortfolio component
  - split into multiple components
  - uses hooks instead of React classes
  - less code better performance (yes this PR adds more code overall but that is caused by tests and strict prop type checking)

### Issues
- using hooks causes console.errors in tests. Enzyme does not support hooks yet, but there are pull requests that should add it shortly: https://github.com/airbnb/enzyme/issues/2011

products-pagination


gmcculloug:
@Hyperkid123 Is the idea to add controls at the top and bottom of the page as Susan mentioned today?

@martinpovolny Please review.

Hyperkid123:
@gmcculloug So the official place for pagination is either top or the bottom. Deepens on the use case. If you can't see the pagination toolbar on the screen (you have to scroll down) then they should be at the top. Susan actually updated mocks and on the card views the pagination is placed on the top toolbar like in the gif. Pagination on the bottom should be mostly on tables and similar components.

PR #116, opened 2019-03-13
Approval workflow

Hyperkid123:
Added approval selection to add/edit portfolio form

### Changes
- migrated to add/edit portfolio form to internal data-driven-forms schema
  - no reason for having the old one
  - if we want to add the image component later on we will have to do it anyway
- added select field with workflows
- removed required parameter from description
- added test coverage for the Modal component

workflows


gmcculloug:
cc @syncrou

src/Helpers/approval/approval-helper.js
```diff
@@ -0,0 +1,3 @@
+import { APPROVAL_API_BASE } from '../../Utilities/Constants';
+
+export const getApprovalWorkflow = () => fetch(`${APPROVAL_API_BASE}/workflows`).then(data => data.json());
```
lgalis: nitpick - plural here: getApprovalWorkflows?

lgalis: Not for this PR - but we should use the generated js client code when that is fixed.

Hyperkid123: > Not for this PR - but we should use the generated js client code when that is fixed.

Yep i know but lately i am constantly switching between client/custom calls because of some missing attributes. I am now planing to do some small "audit" every couple of weeks and replace the custom calls in groups.

PR #121, opened 2019-03-22
Removed initial step from portfolio item ordering dialog.

Hyperkid123:
### Bugs
- Form is not rendering. Turns out that APi returns stringyfied ruby hash instead of JS object.
Screenshot from 2019-03-22 10-01-48
- Issue was reported and i will finish this once we get correct data again. If it takes longer (not fixed before summit), i will replace the API call with a static schema so we have it working for the demo.

### Changes
- removes initial step from order modal

order-modal


gmcculloug:
The  stringyfied ruby hash bug will be fixed with https://github.com/ManageIQ/topological_inventory-api-client-ruby/pull/7

gmcculloug:
@Hyperkid123 Please review travis failures.

PR #127, opened 2019-03-26
Update release cycle with qa

karelhala:
Currently when travis releases new version beta branhces are ignored and app is always pushed to stable folder. This PR fixes such issue and adds correct build steps.


karelhala:
@Fryguy is there any reason for promote script? I removed it and used release script from [RedHatInsights/insights-frontend-builder-common](https://github.com/RedHatInsights/insights-frontend-builder-common).

Fryguy:
@karelhala Because they don't do the same thing, so let's keep it.

The release script pushes the webpack bundle for master into the target branches, but the downside is that there is no corresponding branch in *this* repo.  The promote script actually creates a merge commit on the target branch by merging master into that branch.  Travis will then separately run a CI task for that branch, and at the end it will run the release script.

.travis/custom_release.sh
```diff
@@ -4,10 +4,12 @@ NODE_ENV=production npm run build
...
     .travis/release.sh "ci-beta"
+    .travis/release.sh "ci-beta"
+    
+elif [ "${TRAVIS_BRANCH}" = "stable-master" ]; then
+    .travis/release.sh "qa-stable"
+    .travis/release.sh "ci-stable"
 
-    .travis/promote.sh "qa-beta"
-    .travis/promote.sh "ci-stable"
-    .travis/promote.sh "qa-stable"
```
Fryguy: Please don't change this as what you have is not what we want to do.  Also, regardless of the promote vs release thing, it's just not what we want because a) it changes where we want things to auto-deploy and b) has ci-beta twice.

Fryguy:
The other thing the promote branch does is that it gives a consistent way to manually promote a branch to another branch.  That is, @gmcculloug can use it to push to production by just running the promote script directly.  The release script cannot be run directly because of how the deploy keys are set up for the build repo.

.travis/custom_release.sh
karelhala: But you need to differentiate if you are running things for beta or stable otherwise your beta release will point to stable chrome and that might cause a lot of problems.

karelhala:
Okay, your custom release process is more complicated than I thought. I'm gonna close this PR and create an issue that you should have 2 different build processes one for stable the other one for beta. And when pushing to `ci-beta`, `qa-beta` and `prod-beta` beta build needs to be used and if pushing to `ci-stable`, `qa-stable` and `prod-stable` current build needs to be used.

Fryguy:
@karelhala Can you explain what you mean by beta pointing to stable?  Release wise, the build process doesn't care about beta or stable.  We just happen to release to beta and stable simultaneously at the moment, but they are separate builds.  Is that what the rest of the changes in your PR are doing?

Fryguy:
@karelhala I didn't see the open issue at #130 ...I'll move my question over there.

PR #133, opened 2019-03-27
Add portfolios sharing unsharing

lgalis:
Added a dialog to allow a portfolio to be shared/unshared:
Screenshot from 2019-04-01 16-11-48
Screenshot from 2019-04-01 16-11-38
Screenshot from 2019-04-01 16-11-20


src/helpers/shared/user-login.js
```diff
@@ -27,6 +28,40 @@ export function getUserApi() {
...
+}
+
 export function getWorkflowApi() {
   return workflowApi;
 }
+
+export async function getUserAccess() {
+  let application = 'rbac'; // String | The application name to obtain access for the principal
+  let opts = {
+    //'username': "mkanoor@redhat.com", // String | Unique username of the principal to obtain access for
```
mkanoor: @lgalis can you remove this

src/forms/portfolio-share-form.schema.js
```diff
@@ -0,0 +1,67 @@
...
+      title: 'Invite Group',
+      name: 'new_share',
+      key: '1',
+      fields: [{
+        name: 'group_uuid',
+        component: componentTypes.SELECT,
+        options: rbacGroups
+      }, {
+        name: 'permissions',
+        isRequired: true,
```
Hyperkid123: if you want field to be required you have to add validate specification:
```javascript
import { componentTypes, validatorTypes } from '@data-driven-forms/react-form-renderer';
...
validate: [{
 type: validatorTypes.REQUIRED
}]
```

the `isRequired` key is just a visual prop for PF4 components

src/forms/portfolio-share-form.schema.js
```diff
@@ -0,0 +1,67 @@
...
+
+const groupShareSchema = (groupShareInfo, permissionVerbs) => (
+  {
+    component: 'sub-form',
+    name: `${groupShareInfo.group_name}`,
+    key: `${groupShareInfo.group_name}`,
+    fields: [{
+      name: `${groupShareInfo.group_name}`,
+      label: `${groupShareInfo.group_name}`,
+      isRequired: false,
```
Hyperkid123: no need for this key

src/forms/portfolio-share-form.schema.js
```diff
@@ -0,0 +1,67 @@
...
+      component: componentTypes.SELECT,
+      options: [ ...permissionVerbs, { value: '', label: 'None' }]
+    }]
+  }
+);
+
+export const createPortfolioShareSchema = (shareItems, permissionVerbs) => {
+  let shareInfo = shareItems.items;
```
Hyperkid123: you should use `const` instead of `let` if you are not re-assigning value to variable. I should add linter rule for this.

src/helpers/share/share-helper.js
```diff
@@ -0,0 +1,19 @@
+import { getUserApi } from '../shared/user-login';
+import { CATALOG_API_BASE } from '../../utilities/constants';
+
+const userApi = getUserApi();
+
+export async function getShareInfo(portfolioId) {
+  //return userApi.shareInfo(portfolioId);
```
Hyperkid123: left over comment

src/presentational-components/shared/pf4-select-wrapper.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { FormSelect, FormSelectOption, FormGroup, TextContent, Text, TextVariants } from '@patternfly/react-core';
 
 const createOptions = (options, inputValue, isRequired) => {
-  if (inputValue && isRequired) {
+  if (inputValue) {
```
Hyperkid123: How does this change affect existing selects? There should be one in order form and when adding/editing portfolio

src/redux/actions/portfolio-actions.js
```diff
@@ -107,3 +108,49 @@ export const selectPortfolioItem = (portfolioItem) => ({
...
+    ...portfolioData
+  }),
+  meta: {
+    notifications: {
+      fulfilled: {
+        variant: 'success',
+        title: 'Success updating portfolio',
+        description: 'The portfolio was shared successfully.'
+      },
+      rejected: {
```
Hyperkid123: I would prefer utilizing the default notification mechanism for rejected responses for now because it shows the actual server error. I would guess that eventually the APIs will return some reasonable error messages that can be displayed to users.

src/redux/actions/portfolio-actions.js
```diff
@@ -107,3 +108,49 @@ export const selectPortfolioItem = (portfolioItem) => ({
...
```
Hyperkid123: ^^

src/redux/actions/rbac-actions.js
```diff
@@ -0,0 +1,10 @@
+import { FETCH_RBAC_GROUPS } from '../action-types';
+import { getRbacGroups } from '../../helpers/rbac/rbac-helper';
+
+export const fetchRbacGroups = () => ({
+  type: FETCH_RBAC_GROUPS,
+  payload: getRbacGroups().then(({ data }) => [
```
Hyperkid123: you don't need the square brackets and cloning the array. `.map` already returns new array.

src/redux/actions/share-actions.js
```diff
@@ -0,0 +1,49 @@
...
+export const sharePortfolio = (portfolioData) => ({
+  type: ActionTypes.SHARE_PORTFOLIO,
+  payload: ShareHelper.sharePortfolio({
+    ...portfolioData
+  }),
+  meta: {
+    notifications: {
+      fulfilled: {
+        variant: 'success',
+        title: 'Success updating portfolio',
```
Hyperkid123: I think this should be something like `Success sharing portfolio`.

src/smart-components/portfolio/share-portfolio-modal.js
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  const onSubmit = data =>
+  {
+    let sharePromises = [];
+    if (data.group_uuid && data.permissions) {
+      sharePromises.push(sharePortfolio(data));
+    }
+
+    shareInfo.map(share => {
```
Hyperkid123: `map` is supposed to be used when you want to use create new array from input array. For this use case i would prefer `forEach`

src/smart-components/portfolio/share-portfolio-modal.js
```diff
@@ -0,0 +1,162 @@
...
+              id: portfolioId,
+              permissions: data[share.group_name],
+              group_uuid: share.group_uuid
+            }));
+          }
+        }
+      }
+    });
+
+    Promise.all(sharePromises).then(goBack).then(() => fetchPortfolios());
```
Hyperkid123: I know that `goBack` is being used in some other modals, but i think it would be better to use `history.push(route)`. Go back might bite us when there is no route to go back to (when you open the page directly from URL for instance).

I plan to replace most of the `goBack` calls in the rest of app soon.

src/smart-components/portfolio/share-portfolio-modal.js
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  const onCancel = () => pipe(
+    addNotification({
+      variant: 'warning',
+      title: 'Share portfolio',
+      description: 'Share portfolio was cancelled by the user.'
+    }),
+    goBack()
```
Hyperkid123: ^^

src/test/smart-components/portfolio/portfolio-item-detail/item-detail-info-bar.test.js
```diff
@@ -23,12 +23,12 @@ describe('<ItemDetailInfoBar />', () => {
     };
   });
 
-  it('should render coeectly', () => {
+  it('should render correctly', () => {
```
Hyperkid123: 👍

src/utilities/constants.js
```diff
@@ -5,3 +5,4 @@ export const APPROVAL_API_BASE = `${process.env.BASE_PATH}/approval/v1`;
 // should be based on locale in future. Dont know why we use en-GB format in mocks
 export const dateOptions = [ 'en-GB', { year: 'numeric', month: 'short', day: 'numeric' }];
 
+export const RBAC_API_BASE = `${process.env.BASE_PATH}/rbac/v1`;
```
Hyperkid123: i think you need to specify whole api version `v1.0`. There were some issues with portfolios and approval 
 requests (adding/editing, fetching workflows etc.) were not executed when just the `v1` path was used.

src/presentational-components/shared/pf4-select-wrapper.js
lgalis: @Hyperkid123  - I had to make this change in order for the initial value to be shown.

src/helpers/share/share-helper.js
lgalis: - we should use the client method that is in the comment - I left it there as a marker to swith to - I'll add a TODO

src/redux/actions/portfolio-actions.js
lgalis: this function was moved to the share-actions.js file - I will make this change over in that file.

src/smart-components/portfolio/share-portfolio-modal.js
lgalis: Agree that we have to replace the goBack. I will leave this as is for now, as it is a modal and can be displayed from both the portfolios list a well as an individual portfolio detail page. We can replace it here when we decide on a common way to handle the back path for modals that can return to different pages ( parent)

src/utilities/constants.js
lgalis: +1

lgalis: `Agree, good to know, I think we will, eventually have to use v1, but will switch to v1.0 for now

lgalis:
All feedback addressed

src/presentational-components/shared/pf4-select-wrapper.js
Hyperkid123: Well this is going to break all the screens with select. This is obviously a bug i will check what is going on but we can't just change it like this.

Hyperkid123: What initial value its not showing?

Hyperkid123:
How do we handle when you don't get any groups? Right now it just shows an empty modal:
screenshot-ci foo redhat com-1337-2019 04 04-09-33-17

I think we should add some warning/notification that there no existing groups instead of showing empty modal.

Hyperkid123:
Sorry but I don't think this one was resolved: https://github.com/ManageIQ/catalog-ui/pull/133#discussion_r272049533. The problem is that the initial values do not match. Empty option is added by default here: https://github.com/ManageIQ/catalog-ui/blob/master/src/presentational-components/shared/pf4-select-wrapper.js#L13

https://github.com/ManageIQ/catalog-ui/pull/133/files/7a42d58af46fc857826ec0959a9cc235f89e792a#diff-ab64e77188dd259cd0256972079cf775: this is no resolved. This code `[...data.map(...)]` works but creates 3 arrays instead of one `Array.prototype.map` returns new array anyway. `data.map(...)` is the correct code.

https://github.com/ManageIQ/catalog-ui/pull/133/files/7a42d58af46fc857826ec0959a9cc235f89e792a#diff-fb2187d73d464fdeb11eec6457b47760: you agreed but there is still just `v1` instead of `v1.0`

There are more unresolved comments like that above.

lgalis:
@Hyperkid123 - regarding the versions - I updated that to v1.0 for the approval for now. For the RBAC, we have used only v1 everywhere - so this can stay as is. ( in fact, this does not work with v1.0)

src/redux/actions/portfolio-actions.js
lgalis: this function was moved to the share-actions.js file - I will make this change over in that file.

src/presentational-components/shared/pf4-select-wrapper.js
lgalis: Removed the changes made to to this file.

src/utilities/constants.js
lgalis: The approval was changed to v1.0 - the rbac only works with v1, this has to stay as is

src/redux/actions/rbac-actions.js
lgalis: I missed this one - changed now.

lgalis:
Addressed remaining feedback and added a message when no groups are available.

PR #134, opened 2019-03-28
Update adding items to portfolio

Hyperkid123:
### Changes
- Created PF4 checkbox (not actual PF component because it does not exist yet)
screenshot-ci foo redhat com-1337-2019 03 28-14-19-51
screenshot-ci foo redhat com-1337-2019 03 28-14-20-01
- Automatically selects first source when adding portfolio items
add-to-portfolio
- Added the option to add custom empty state component to content gallery

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1012,9 +1012,9 @@
       }
     },
     "@patternfly/react-icons": {
-      "version": "3.6.1",
-      "resolved": "https://registry.npmjs.org/@patternfly/react-icons/-/react-icons-3.6.1.tgz",
-      "integrity": "sha512-7PkP7NC3sDmsFAnQIGl+9p14vgx7gj0TZSwtL1ABkAEV9tXcEjWsyyZP2fahfTKVE9aPrOZ/YINsr8+iGCOTnA=="
+      "version": "3.7.0",
+      "resolved": "https://registry.npmjs.org/@patternfly/react-icons/-/react-icons-3.7.0.tgz",
+      "integrity": "sha512-ep1ImwHL/Rzd4p66kZKmWV97aZDe967AHD70qTrwoNU1xELSD2bClckR4P/RK9NY3DVuqkUwGNYFICfcgdpztQ=="
     },
     "@patternfly/react-styles": {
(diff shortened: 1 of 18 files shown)


gmcculloug:
@Hyperkid123 Why they change to "Automatically selects first source when adding portfolio items"?

I could see this if only one source existed.  But if there are more and that's not the source the user wanted they now have to wait for the UI to load the resources before switching.

Hyperkid123:
I did that because we don't have any empty state design yet. Right now you will get into situation where when the screen will tell you that no items were found if you did not select any source. 

And users actually don't have to wait for the services to be loaded. They have to wait for sources to be loaded (same as before). They can switch even while services are loading. Services have separate storage based on the platform so you don't have to worry about any asynchronous shenanigans.

But if this does not make any sense to you i can remove it.

jonnyfiveiq:
Ok so first the platform selection should be first in the row, then followed by search field.
If there is only one platform then it should default to that.
If there are no platforms then we should get emptystate with button link to topology add source wiz.
If there are 2 or more platforms in the list and none are default then it should provide emptystate pointing to select a platform.
The platform should have a check box where by you can set it as default so that whenever that user goes to platforms it selects that first.

decide what for pre/post mvp.

Hyperkid123:
> If there are no platforms then we should get emptystate with button link to topology add source wiz.

This is probably not a MVP yet because o know that the topology team wants to share their implementation of adding source dialog but probably not before summit. Cost management and one other service has this requirement too, but that means that the topology must be a globally accessible from chrome. Right now it lives under the HCM section.

The rest we can do easily i think. @gmcculloug Any thoughts on this?

sbuenafe-rh:
@Hyperkid123 for the PF checkbox component, have you seen https://v2.patternfly.org/documentation/react/components/checkbox

Hyperkid123:
@sbuenafe-rh yep but you have something different in the mocks.

sbuenafe-rh:
@Hyperkid123 you should pull from that PF4 component, I'll have to eventually update my mocks with PF's checkbox.

PR #143, opened 2019-04-02
Fixed wrong query param for service offerings.

Hyperkid123:
Topology API introduced new filtering queries. The service offerings query has to be updated to return some results.


martinpovolny:
I cannot help a growing feeling that we are merging a number of PRs where the UI is dealing with incompatible changes in the API. 

I wonder how is the schema versioning working and if there are tests for the API usecases that the UI is consuming.

PR #145, opened 2019-04-05
Portfolio items icons

Hyperkid123:
### Changes
- added icons to portfolio items
- added icons to portfolio item detail
- added source icon to service if no icon is specified
  - for now it defaults to openshift icon but will be changing based on provider once we have more.
- added lazy loading to images to increase performance
  - network request for image will be fired only when the image is in view port

#### Lazy loading
lazy-loading

#### Item detail
screenshot-ci foo redhat com-1337-2019 04 05-11-01-32
#### Default icon if icon is missing
screenshot-ci foo redhat com-1337-2019 04 05-11-01-43

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10083,6 +10083,11 @@
       "integrity": "sha1-4j8/nE+Pvd6HJSnBBxhXoIblzO8=",
       "dev": true
     },
+    "lodash.debounce": {
+      "version": "4.0.8",
+      "resolved": "https://registry.npmjs.org/lodash.debounce/-/lodash.debounce-4.0.8.tgz",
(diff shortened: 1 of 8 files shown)


gmcculloug:
cc @syncrou @lindgrenj6

gmcculloug:
@Hyperkid123 This is deployed and looks great.  I think we just need one followup to update the icon and display name on the order dialog screen:

image

Hyperkid123:
Yeah i agree. I would love to talk with Susan about the order modal a bit. Every time i open it i like it less and less :D.

gmcculloug:
Sounds great, definitely talk to @sbuenafe-rh about it.  The initial delay while waiting for the dialog to generate could use some cosmetic help. 🙂 I'll try to get a screen grab of it, but it can be pretty quick some times.

PR #150, opened 2019-04-09
Order items

Hyperkid123:
### Changes
- added redux actions that links orders, order items, requests
- added new data to order items

### TBD
- pagination and filtering
  - i would suggest doing this in a follow up PR. We need to figure out how we want to paginate and filter because with current API support it does not make sense because its more efficient to pull everything at once than loading data based on orders (too many requests).
- correct progress messages
  - right now there are only two states finished/pending based on order item/request/order states.
  - also each entity has different flag for same state.

Peek 2019-04-11 14-05


gmcculloug:
cc @syncrou

PR #151, opened 2019-04-10
Fix card line clamping in Firefox.

Hyperkid123:
I've replaced the CSS with react component. Its not pretty but if we want multi line ellipsis effect there is no other way unfortunately. Firefox simply does not support `line-clamp` CSS property.

### Before
Screenshot from 2019-04-10 14-39-22

### After
Screenshot from 2019-04-10 14-41-29

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7074,14 +7074,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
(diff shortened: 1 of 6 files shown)


gmcculloug:
@Hyperkid123 Please add some before/after screenshots.

cc @sbuenafe-rh

PR #152, opened 2019-04-10
Disabled product edit page.

Hyperkid123:
Disabled editing of portfolio item.

### Changes
- Commented routes and links that lead to edit page in UI (don't want to delete it because it will come back probably)
- Added redirect to the edit page if its accessed directly via URL.
- Hide the approval data from detail page.


Hyperkid123:
@lgalis ah i commented event the order route... nice catch

Hyperkid123:
@lgalis ok modal is back and i've added test so we wont make same mistake again.

PR #154, opened 2019-04-11
Remove link from Settings text in empty state.

gmcculloug:
cc @sbuenafe-rh


src/smart-components/platform/platforms.js
```diff
@@ -46,7 +46,7 @@ class Platforms extends Component {
           </Text>
           <Text component={ TextVariants.p }>
             To connect to a source, go to <a href={ `${document.baseURI}hybrid/settings/catalog-sources` }>Catalog sources</a>&nbsp;
-            under <a href="javascript:void(0)">Settings</a>
+            under Settings</a>
```
lgalis: the </a> needs to be removed as well

PR #163, opened 2019-04-18
Added redux action creator for remove items from portfolio.

Hyperkid123:
### Bug fixes
- adds notifications after portfolio items were removed successfully

### Changes
- added redux action creator for remove items from portfolio
  - before the async function was called directly so it did not allow to specify notifications when resolved.


src/redux/actions/portfolio-actions.js
```diff
@@ -107,3 +107,17 @@ export const selectPortfolioItem = (portfolioItem) => ({
...
 
+export const removeProductsFromPortfolio = portfolioItems => ({
+  type: ActionTypes.REMOVE_PORTFOLIO_ITEMS,
+  payload: PortfolioHelper.removePortfolioItems(portfolioItems),
+  meta: {
+    notifications: {
+      fulfilled: {
+        variant: 'success',
+        title: 'Products removed',
+        description: 'You have removed three products from the Global portfolio. Undo this if it was a mistake.'
```
lgalis: The 'three' word here needs to be removed/replaced. Either use the actual number of products removed or a general term.

Hyperkid123: Thanks i just copied it from mocks :D

Hyperkid123: Ok now it will print the number. We can talk about some text transformation tools later.

src/redux/actions/portfolio-actions.js
```diff
@@ -107,3 +107,17 @@ export const selectPortfolioItem = (portfolioItem) => ({
...
 
+export const removeProductsFromPortfolio = (portfolioItems, portfolioName) => ({
+  type: ActionTypes.REMOVE_PORTFOLIO_ITEMS,
+  payload: PortfolioHelper.removePortfolioItems(portfolioItems),
+  meta: {
+    notifications: {
+      fulfilled: {
+        variant: 'success',
+        title: 'Products removed',
+        description: `You have removed ${portfolioItems.length} product${portfolioItems.length !== 1 ? 's' : ''} from the ${portfolioName} portfolio. Undo this if it was a mistake.` // eslint-disable-line max-len
```
martinpovolny: oh well, this is an ugly hack that is not i18n compatible at all

we should probably start doing something about the i18n at some point

Hyperkid123: You are right. I am thinking more and more about react intl (https://github.com/yahoo/react-intl). I will definitely do more research after summit.

But i think for i18n there should be some platform level agreement because a lots of stuff like error messages must be also supported by the APIs. Or take the sources for instace. You have form schema stored in DB but it must be translated correctly in different apps owned by different teams.

If we start using something without letting anybody else know, we could end up using x amount of different i18n tools per app because of disparities between services.

martinpovolny: react-intl looks pretty good

I am not aware of any directions or solution in the platform. At the same time I suspect that some people expect/suppose that the apps are being developed as properly internationalized.

We should really start looking into this rather sooner than later.

lgalis:
@Hyperkid123  - Looks good for now - I'll merge once the travis build issue is fixed.

Hyperkid123:
@lgalis ok tests are fixed now.

Issue #177, opened 2019-04-24
Time missing in the Order Stages


mkanoor:
Only date is visible in the Order Stages even though the heading says "Date & Time"

![Screen Shot 2019-04-24 at 5 38 24 PM](https://user-images.githubusercontent.com/6452699/56695880-2d356380-66b8-11e9-9308-ac90f8205818.png)

gmcculloug:
@sbuenafe-rh I recall you said this should be relative date/time and on hover over it should display UTC.  Is that still the plan?

sbuenafe-rh:
@gmcculloug yes, the guidance is relative date/time with tooltip with precise date and UTC time. We do display the relative time for when it was ordered, and when the order was updated in the row. I just brought forward the precise date/time for the steps -- in thinking forward that our event log will have the precise timestamps.

PR #180, opened 2019-04-25
Added time and UTC tool tip for order relative dates.

Hyperkid123:
### Changes
- Added time to date time to relative order dates.
- Added time to the date time column

fixes: #177 

Screenshot from 2019-04-30 14-00-34

cc @sbuenafe-rh

### DIFF
--- a/src/helpers/shared/helpers.js
+++ b/src/helpers/shared/helpers.js
@@ -38,4 +38,4 @@ export const createUpdatedLabel = orderItem => {
   return `Updated ${orderedAgo} ${orderedAgo > 1 ? 'days' : 'day'} ago`;
 };
 
-export const createDateString = date => moment(new Date(date), 'DD-MMM-YYYY').format('DD MMM YYYY');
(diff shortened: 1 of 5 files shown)


Hyperkid123:
@sbuenafe-rh is this the correct visual change? I am kind of guessing with this. The tool tip is PF4 tool tip component.

sbuenafe-rh:
@Hyperkid123 looks good, but you don't need the tooltip on the date/time because it's already precise date/time. 

as a side note, you will need the tooltip showing the precise date/tip over the relative times (ie Updated 2 days ago)

Hyperkid123:
@sbuenafe-rh ok i am quire confused by the #177 conversation because the date is not standard UTC format (the in the tool tip is). Should i swap the format and remove the tool tip?

sbuenafe-rh:
@Hyperkid123 you can take out the day "Thu" and the seconds from the time format and you'll be set for the tooltip.

Hyperkid123:
Ok should be correct now. Thanks @sbuenafe-rh

Issue #182, opened 2019-04-25
UI hangs when a non privileged user tries to "Delete" Portfolio


mkanoor:
When a non privileged user tries to delete a Portfolio the API returns a 403 forbidden but the UI hangs.
Maybe similar to #181

Hyperkid123:
Resolved by #184

Issue #181, opened 2019-04-25
UI hangs when a non privileged user "Creates Portfolio"


mkanoor:
If a user doesn't have access to create Portfolios the REST API returns a 403 Forbidden
```
HTTP/1.1 403 Forbidden

{"message":"Write access not authorized for Portfolio"}
```

This causes the UI to keep spinning.
If you switch to a privileged user you can create the portfolio successfully.

Hyperkid123:
Resolved by #184

PR #184, opened 2019-04-29
Added global redux rejected loading handler.

Hyperkid123:
### Changes
- New redux middleware for rejected actions
  - is triggered by rejected action creator
  - sets loading state to false in redux store
  - UI no longer "spins" after unsuccessful requests
- Added new key to response messages (data.message)

solves: #181 and #182


Hyperkid123:
@mkanoor can you please test if it fixed problems in #181 and #182? This should cover everything but i would if someone double checked it.

lgalis:
@Hyperkid123 - I still see the infinite spinner - the error occurs in the add notification for the error.

Hyperkid123:
@lgalis can you PM me steps to reproduce? I was testing in with 401 and 5xx errors and it was working for those.

Hyperkid123:
@lg ok the error should be fixed now. There was error key missing in the notifications middleware config.

PR #188, opened 2019-04-30
Added custom render trigger on order item component.

Hyperkid123:
### Changes
- added custom trigger for OrderItem render method
  - Expanding/Collapsing the order item was slow
  - When expanding every single order item was re-rendered
  - added direct connection to redux from each order item
  - added should component update trigger 
  - render times are down from 125 ms to 15 ms per order item

### DIFF
--- a/src/smart-components/order/order-item.js
+++ b/src/smart-components/order/order-item.js
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
 import {
   DataListCell,
   DataListContent,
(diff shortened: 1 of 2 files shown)


gmcculloug:
Based on the description 👍 
@martinpovolny Please review

src/smart-components/order/order-item.js
```diff
@@ -27,91 +28,100 @@ import { getOrderIcon, getOrderPortfolioName } from '../../helpers/shared/orders
 import { createOrderedLabel, createUpdatedLabel, createDateString } from '../../helpers/shared/helpers';
 import createOrderRow from './create-order-row';
 
-const OrderItem = ({ item, isExpanded, handleDataItemToggle, portfolioItems }) => {
-  const { finishedSteps, steps } = createOrderRow(item);
-  const orderedAt = createOrderedLabel(new Date(item.ordered_at));
-  const updatedAt = createUpdatedLabel(item.orderItems);
+class OrderItem extends Component {
+  shouldComponentUpdate({ isExpanded }) {
```
lgalis: this could further be improved by returning true only when the item is being expanded, not when closed.

martinpovolny: I do not follow.

martinpovolny:
Looks good to me. Please, resolve the conflicts.

src/smart-components/order/order-item.js
Hyperkid123: If i understand your suggestion correctly it would mean the component would never close.

PR #201, opened 2019-05-10
Open the 'Manage product' link in a new tab

lgalis:
Open the 'Manage product' link in a new tab

Solves https://projects.engineering.redhat.com/browse/SSP-383

### DIFF
--- a/src/test/smart-components/order/orders.test.js
+++ b/src/test/smart-components/order/orders.test.js
@@ -6,7 +6,7 @@ import configureStore from 'redux-mock-store' ;
 import { shallowToJson } from 'enzyme-to-json';
 import { MemoryRouter } from 'react-router-dom';
 import promiseMiddleware from 'redux-promise-middleware';
-import { DataList, DataListContent, DataListCell } from '@patternfly/react-core';
+import { DataList, DataListContent } from '@patternfly/react-core';
 
 import Orders from '../../../smart-components/order/orders';
 import { orderInitialState } from '../../../redux/reducers/order-reducer';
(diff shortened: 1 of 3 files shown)


src/test/smart-components/order/orders.test.js
```diff
@@ -85,11 +98,39 @@ describe('<Orders />', () => {
...
+    apiClientMock.get(`${CATALOG_API_BASE}/orders`, mockOnce({ body: { data: []}}));
+    apiClientMock.get(`${APPROVAL_API_BASE}/requests`, mockOnce({ body: { data: []}}));
+    apiClientMock.get(`${CATALOG_API_BASE}/portfolio_items`, mockOnce({ body: { data: []}}));
+    apiClientMock.get(`${CATALOG_API_BASE}/order_items`, mockOnce({ body: { data: []}}));
+
+    const wrapper = mount(<ComponentWrapper store={ store }><Orders { ...initialProps } linkedOrders /></ComponentWrapper>);
+    setImmediate(() => {
+      wrapper.update();
+      expect(wrapper.find(DataListContent).first().props().isHidden).toEqual(true);
+      expect(wrapper.find(DataListCell).last().props().children.props.children.props).toMatchObject(
```
Hyperkid123: You can also call `find` on enzyme nodes so you dont have to chain the children calls. Something like
```javascript
expect(wrapper.find(DataListCell).last().find('a').props()).to....
```
It should return the same result.

lgalis: done

lgalis:
@Hyperkid123 - updated, thank you!

PR #204, opened 2019-05-14
Toolbar renderer

Hyperkid123:
### Changes
- added new toolbar renderer which takes JSON schema of a toolbar and renders PF component.
  - it should make toolbars easier to create and more consistent across the UI
- replaced portfolios-toolbar and portfolio-filter-toolbar with the toolbar renderer
 - more will come in different PRs
- needed to update PF4 components because Dropdown was not working. That messed up some snapshots.
- no "visible" UI changes

### TO DO
- [x] add helpers for common toolbar structures
  - [x] toolbar group -> toolbar item -> action
  - [x] router link -> button
- [x] create common toolbar components
  - [x] filter input


src/test/presentational-components/portfolio/__snapshots__/portfolios-filter-toolbar.test.js.snap
```diff
@@ -1,47 +1,79 @@
...
-          component="button"
-          isActive={false}
-          isBlock={false}
-          isDisabled={false}
-          isFocus={false}
-          isHover={false}
-          type="button"
-          variant="primary"
+        <Component
+          component="ToolbarItem"
```
lgalis: should this field be named something other 'component' - maybe 'type'?

Hyperkid123: So this is just a generic component provider which holds the Context with component list that returns the actual React Component that should be rendered. You can actually reference React component directly instead of the string and that is what is happening in virtual dom. Yout don't see the string but React component from mapper. Unfortunately Jest is not smart enough to map the string to component into snapshots.

lgalis: ok, sounds good

PR #209, opened 2019-05-16
Added react intl provider

Hyperkid123:
### Changes
- adds react-intl context to App (https://github.com/formatjs/react-intl)
- we can now use proper pluralization in messages
- also it prepares  the app for the i18n

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
-import React, { Component } from 'react';
+import React, { Component, Fragment } from 'react';
 import { withRouter } from 'react-router-dom';
 import { Grid, GridItem } from '@patternfly/react-core';
 import { Main } from '@redhat-cloud-services/frontend-components';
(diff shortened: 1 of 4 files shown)


src/App.js
```diff
@@ -70,16 +73,18 @@ class App extends Component {
...
-        <Main style={ { marginLeft: 0, padding: 0 } }>
-          <Grid style={ { minHeight: MIN_SCREEN_HEIGHT } }>
-            <GridItem sm={ 12 }>
-              <Routes childProps={ this.props } />
-            </GridItem>
-          </Grid>
-        </Main>
-      </React.Fragment>
+      <IntlProvider locale="en">
+        <Fragment>
```
martinpovolny: Do you need the `Fragment` inside the `IntlProvider`?

Hyperkid123: You can have only one child in the provider. (Same as Router used to be)

martinpovolny: I see. I just hit into that trying to do the same in Source.

PR #210, opened 2019-05-20
Fix invalid DOM nesting in portfolio card.

Hyperkid123:
### Bug Fixes
- fixed invalid DOM nesting in portfolio card
  - whole component was wrapped in Link and the header has dropdown with links
  - it was causing invalid DOM nesting console error
  - moved the portfolio card link only to card body.

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -187,3 +187,11 @@
     border-top-color: var(--pf-c-button--disabled--BackgroundColor);
   }
 }
+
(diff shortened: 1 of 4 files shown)


Hyperkid123:
> the body of the cards can be small and a large portion of the card will now not be clickable

I the link has 100% height of its parent. I've also removed default card-body spacing and moved it to the link so event the edges are clickable. It should be working

PR #212, opened 2019-05-21
Add undo portfolio delete

Hyperkid123:
marge after client update: https://github.com/RedHatInsights/javascript-clients/pull/18

### Changes
- adds undo action after deleting portfolio item

undo-portfolio-item


sbuenafe-rh:
Looks good. Only nit picky comment is that the message for restoring products should be:

"Products have been restored"

If we are referring to the portfolio items as products or product items we should be consistent.

Hyperkid123:
> Looks good. Only nit picky comment is that the message for restoring products should be:
> 
> "Products have been restored"
> 
> If we are referring to the portfolio items as products or product items we should be consistent.

Thanks. Changes is here: https://github.com/ManageIQ/catalog-ui/pull/212/files#diff-f18673ef95a64a3e26c163130463b658R124

i am to lazy to make another gif.

PR #219, opened 2019-06-18
Added copy portfolio item feature.

Hyperkid123:
@syncrou @gmcculloug I have a question about the new portfolio name. UI renders the `display_name` attribute of portfolio item because that is the human readable equivalent of `name` but the copy endpoint changes the `name` attribute so the `display_name` is unchanged. 

I belive the `display_name` should be modified instead. Any thoughs?
(endpoint PR: https://github.com/ManageIQ/catalog-api/pull/289)

### Changes
- added copy portfolio item modal
- mocks: https://docs.google.com/presentation/d/1Z0Jd5JQa1eavch_ZbTJ8YkDjRZpu6twyece1T-k8xOw/edit#slide=id.g5b73dafedf_0_47
copy-portfolio-item

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -900,21 +900,22 @@
       }
     },
     "@data-driven-forms/pf4-component-mapper": {
-      "version": "1.9.11",
-      "resolved": "https://registry.npmjs.org/@data-driven-forms/pf4-component-mapper/-/pf4-component-mapper-1.9.11.tgz",
-      "integrity": "sha512-/4AcxhcdetGxZK+gpYVv3RnAgyy1lYdsyB80ddKEN5PZ5SBZh6D1+3B+5JEfqych+86SNBSlV60SPOUJThwOgQ==",
+      "version": "1.11.0",
+      "resolved": "https://registry.npmjs.org/@data-driven-forms/pf4-component-mapper/-/pf4-component-mapper-1.11.0.tgz",
+      "integrity": "sha512-IZzmijukSHbRI7X6Pj+jRjUHTOtzBsMStuWgDhgD0WEVwe+WhmFTuRpgmXdDbdKWoOha4BY2CyvMNTNrk/OiRQ==",
       "requires": {
         "@patternfly/patternfly-next": "^1.0.87"
       }
     },
     "@data-driven-forms/react-form-renderer": {
-      "version": "1.9.11",
-      "resolved": "https://registry.npmjs.org/@data-driven-forms/react-form-renderer/-/react-form-renderer-1.9.11.tgz",
-      "integrity": "sha512-oORV3BCUVszlqhi3oe8VFCIZHImI7InBm8qJy0KMnohie2IMwYTXwDSSzB7pD92NsI3sOf1PzcP/3l9NSZRPHw==",
+      "version": "1.11.0",
+      "resolved": "https://registry.npmjs.org/@data-driven-forms/react-form-renderer/-/react-form-renderer-1.11.0.tgz",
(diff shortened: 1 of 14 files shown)


lgalis:
Looks good - I have one question: When the portfolio item is copied into a different portfolio, and that portfolio already contains an item with the same name, there will be a duplicate name there - cannot recall if duplicate names are allowed on portfolio items.

syncrou:
@Hyperkid123 @lgalis I'm looking into both your questions

syncrou:
@lgalis @Hyperkid123 we are working on adding a GET endpoint to pull the next available name, and updating both `name` and `display_name`.

Hyperkid123:
> LGTM, but would like to wait for the `display_name` / `name` JIRA: https://projects.engineering.redhat.com/browse/SSP-511
> and GET /portfolio_items/:id/... endpoint JIRA: https://projects.engineering.redhat.com/browse/SSP-512 tickets to be complete before merging this.

I am OK with this just let us know when it will be ready

gmcculloug:
@Hyperkid123 The PRs for the Jira stories referenced above are merged.

https://github.com/ManageIQ/catalog-api/pull/314 and https://github.com/ManageIQ/catalog-api/pull/315

Hyperkid123:
@gmcculloug @syncrou @lgalis i've added API calls to secure unique copy name.

Hyperkid123:
> I found a few issues:
> 
> 1. When doing a copy of a portfolio item, using the breadcrumbs to navigate to the portfolio - it does not show the newly created copy
> 2. On the copy page - the select component that lists the target portfolio - once modified from the original value, will display the same value ( incorrect) for any selection. ( Engineering 1 in my test).

1. This was occurring only when copping to the same portfolio - added data reload if that si the case.
2. I was sending wrong value on the on change action to the form.

@lgalis  both issues are fixed, thanks for your testing!

lgalis:
@Hyperkid123 - thank you for fixing the 2 issues I mentioned.
One more problem I see is the fact that the 'Copy of' is added twice to the name each time a portfolio item is copied: when the copy modal opens, there is one 'Copy of', but a second is added on Save:

Screenshot from 2019-07-02 09-00-52
Screenshot from 2019-07-02 09-01-03
Screenshot from 2019-07-02 09-01-21

Hyperkid123:
@lgalis i think i know what is going on here. So i am pulling the name from the name provided endpoint and sending that name to the copy endpoint which is adding another `Copy of` string to it.

@syncrou is this intentional? I was expecting that when i send name to the copy endpoint it will use that and when i don't send any it will generate `Copy of something`. This might be potentially an issue if we ever allow users to choose their own name.

lgalis:
@Hyperkid123 , @syncrou  - I think the copy endpoint should not change the name provided for the copy, just fail if the name already exists. Maybe a separate endpoint can be added that does not accept a new name can be created - and this endpoint would always use the next name.

Hyperkid123:
>  and this endpoint would always use the next name.

@lgalis current endpoint does that if you don't send any name to it i think. But we are sending name that is provided by the backend which is already generated as next unique name but is prefixed again.

lindgrenj6:
@Hyperkid123 @lgalis I think I know what's happening here.
The `next_name` endpoint gets queried when displaying the modal, then when it gets submitted we're doing name resolution again, and since the `Copy of ###` name doesn't exist, we tack `Copy of` on the front of the string again. I'll just make it return an un-altered string if there isn't one like it already. I'll get that fixed in the backend pronto!

lgalis:
Looks good - verified that all issues were addressed.

PR #220, opened 2019-06-19
Added disabled state to portfolio items in content gallery.

Hyperkid123:
### Changes
- switched order of async calls when removing portfolio items
  - undo notification is not showed until items are removed, list is updated and rendered
  - portfolio items list update is now part of the remove action instead of being called after remove is finished
- added state mutation on portfolio items remove pending/success
  - this causes the add/remove items buttons to be disabled correctly
- added disabled state to portfolio items while item is being removed
  - adds gray overlay
  - disables events on card (user can no longer redirect to item which is being deleted)

disable item in progress

cc @sbuenafe-rh

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -188,6 +188,20 @@
   }
 }
 
+.portfolio-item-progress {
+  position: relative;
+  height: 330px;
+  .content-gallery-card {
+    position: absolute;
+    width: 100%;
+    &.progress-overlay {
+      background-color: rgba(210, 210, 210, .5);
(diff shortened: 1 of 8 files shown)


lgalis:
@sbuenafe - please review

sbuenafe-rh:
@Hyperkid123 looks good per our convo this morning. I'll look into the disabled selected card tile vs. disable all cards a little more. Just to note: Remove products flow will change slightly in the future once the PF card view demo is released.

PR #222, opened 2019-07-08
Move remove portfolio item action

Hyperkid123:
https://marvelapp.com/625jh26/screen/58846717

### Changes
- portfolio items can be now selected directly in portfolio item screen
- remove portfolio items actions no longer has its own route
- remove portfolio items is not an action on portfolio toolbar

### Before
remove-pi-be

### After
remove-portfolio-items


gmcculloug:
@Hyperkid123 In the `After` image the "Add products" button is greyed-out when you first enter the Portfolio.  I would expect that it would be enabled then and greyed-out once you start selecting Products.  It looks like it is doing the opposite.

Hyperkid123:
> @Hyperkid123 In the `After` image the "Add products" button is greyed-out when you first enter the Portfolio. I would expect that it would be enabled then and greyed-out once you start selecting Products. It looks like it is doing the opposite.

@gmcculloug is greyed-out because the data was not loaded yet. We can disable it when some items are select. I will get in touch with Susan.

gmcculloug:
@Hyperkid123 I think I was just seeing the delay so I don't have an issue with it now, thanks.

PR #224, opened 2019-07-09
Migrate portfolio to function

Hyperkid123:
### Changes
- no visual changes
- split Portfolio into multiple independent components
- Portfolio is now function component using hooks instead of class component

### Fixes
- fixed a small issue when portfolio shows an empty state for a split second before rendering portfolio items
- portfolio items now properly shows loading state when switching between portfolios


src/smart-components/portfolio/portfolio-products.js
```diff
@@ -0,0 +1,74 @@
...
+import ToolbarRenderer from '../../toolbar/toolbar-renderer';
+import createPortfolioToolbarSchema from '../../toolbar/schemas/portfolio-toolbar.schema';
+import AddPortfolioModal from './add-portfolio-modal';
+import RemovePortfolioModal from './remove-portfolio-modal';
+import SharePortfolioModal from './share-portfolio-modal';
+import OrderModal from '../common/order-modal';
+import PortfolioEmptyState from './portfolio-empty-state';
+import ContentGallery from '../content-gallery/content-gallery';
+
+const PortolioProducts = ({
```
lgalis: I think we should try to use the 'portfolio item' term, not 'product' everywhere possible - except the strings that are actually displayed on the screen.

lgalis:
@Hyperkid123 - looks good, only one issue:
on the Add Products page - the Filter is not displaying the input, but is filtering correctly:
Screenshot from 2019-07-09 11-54-19

Hyperkid123:
@lgalis filtering should be OK

lgalis:
@Hyperkid123 - thank you - looks good, merging.

PR #225, opened 2019-07-09
Modified order modal to accept different UI schema types.

Hyperkid123:
~~Merge only after https://github.com/data-driven-forms/react-forms/pull/26. There was a bug with argument assignment to one of the validator types.~~

### Changes
- no visual changes
- added simple function that creates correct schema for order modal form
  - there is no explicit way to identify schema from Openshift, that should be added
- now we can order items from both openshift and Tower

### DIFF
--- a/src/smart-components/order/order-service-form-step-configuration.js
+++ b/src/smart-components/order/order-service-form-step-configuration.js
@@ -3,10 +3,29 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router-dom';
 import { Bullseye, Radio, Form, Title, Stack, StackItem } from '@patternfly/react-core';
+import { schemaParsers } from '@data-driven-forms/react-form-renderer';
 
 import FormRenderer from '../common/form-renderer';
 import { fetchServicePlans, sendSubmitOrder } from '../../redux/actions/order-actions';
 import { fetchProviderControlParameters } from '../../helpers/portfolio/portfolio-helper';
+
+/**
+ * TO DO
+ * We need some explicit way of telling what kind of schema is being received
+ */
+const createFormSchema = (servicePlanSchema, providerControlParameters) => {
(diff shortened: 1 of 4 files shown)


src/smart-components/order/order-service-form-step-configuration.js
```diff
@@ -3,11 +3,30 @@ import PropTypes from 'prop-types';
...
 import FormRenderer from '../common/form-renderer';
 import { fetchServicePlans, sendSubmitOrder } from '../../redux/actions/order-actions';
 import { fetchProviderControlParameters } from '../../helpers/portfolio/portfolio-helper';
 
+/**
+ * TO DO
+ * We need some explicit way of telling what kind of schema is being received
+ */
+const createFormSchema = (servicePlanSchema, providerControlParameters) => {
+  const providerFields = schemaParsers.mozillaParser(providerControlParameters).schema.fields
```
lgalis: If we allow a schema type for the servicePlanSchema, we should do the same for the providerControlParameters , rather than expect these always to be Mozilla type. We could either: 

1. Allow a schema type to be specified for the providerControlParameters 
or
2. Assume the same schema type for the providerControlParameters as the one specified for the servicePlanSchema, with mozzilla as default.

Hyperkid123: I agree we need to get more details about this. That is why i added the TO DO. I expect i'll be revisiting this piece of code very soon.

gmcculloug: If it helps, I don't think we need the Control Parameters for this initially pass of Tower support.  We can change providerControlParameters to be whatever you need, but the end result is it will not require anything additional to be displayed.

lgalis: Understood - was making a general point, not specific to any type of provider.

lgalis: We can merge this as is for now, and revisit later.

PR #233, opened 2019-07-25
Optimistic ui for portfolios components.

Hyperkid123:


Hyperkid123:
cc @sbuenafe-rh

Hyperkid123:
@lgalis we have been talking about this with Susan a couple of weeks ago. We should be using it instead of loaders and spinners which can take over the screen very easily.

What we don't know just yet, is how to show cards in disabled state. That is probably a placeholder for now.

sbuenafe-rh:
Looks good! Will look into other options for the show cards in disabled state.

PR #235, opened 2019-07-26
Hide empty attributes in product detail page.

Hyperkid123:
- empty attributes are no longer showing on portfolio item detail
- removed sample repository link because it was always a dead link
- removed duplicate description text and move overview section to the top

screenshot-ci foo redhat com-1337-2019 07 26-09-36-08


sbuenafe-rh:
Looks good!

PR #236, opened 2019-07-30
Change share portfolio action flow.

Hyperkid123:
### Changes
- using select from data-driven forms
  - should fix tower multi select component
- switched action flow of share portfolio action
  - user is redirected before the action is finished
  - this eliminates the delay and need for spinners
  - no visual changes, UI just responds immediately
- updated version of data driven forms
  - this automatically disables submit button if onSubmit is promise in progress


Hyperkid123:
@gmcculloug @sbuenafe-rh I think we should make some changes to the share modal. The whole thing feels and acts a bit strange to me.

gmcculloug:
@Hyperkid123 Let's discuss your thoughts on the sharing dialog after SU.

PR #246, opened 2019-08-22
Filter sources by application type

Hyperkid123:
### Changes
- we only query sources that are of `/insights/platform/catalog` application type
- added graphql API instance
  - added interceptor to catch graphql error response and throw it as an actual error

Now we can see only the Tower source (in both platforms and when adding items to portfolios):
screenshot-ci foo redhat com-1337-2019 08 22-15-03-27


gmcculloug:
@Hyperkid123 Overall this looks good.  Will the Platforms also be filtered when going through "Add Product" from the Portfolio screen?

image

Hyperkid123:
@gmcculloug yes it all uses the same function to get the data

Hyperkid123:
@lgalis attribute was added thanks for noticing!

PR #249, opened 2019-08-26
Add cancel order button

Hyperkid123:
### Changes
- updated pf
  - needed for new modal features
  - required update of several snapshots
- added cancel order to orders

screenshot-ci foo redhat com-1337-2019 08 26-11-23-47

screenshot-ci foo redhat com-1337-2019 08 26-11-24-23

PR #252, opened 2019-08-27
[WIP] Rework orders

Hyperkid123:
waiting: #249 

description: TODO


Hyperkid123:
@lgalis yeah the whole filtering is just a placeholder. Orders do not have a name now so its kind of weird right now.

I will fix the wrong redirect thanks for that.

PR #255, opened 2019-09-09
Enabled editing of portfolio item attributes.

Hyperkid123:
### Changes
- added form to edit portfolio item attributes

edit-portfolio-item


gmcculloug:
cc @sbuenafe-rh

sbuenafe-rh:
@Hyperkid123 Looks good, should one of those fields be for the image uploader too?

gmcculloug:
I would expect image uploader to be a separate PR.

Hyperkid123:
@sbuenafe-rh i was thinking we will do the same pattern as GH does. When entering the edit page the will display some button/icon button which will serve as the uploader. What do you think about that?

sbuenafe-rh:
@Hyperkid123 GH pattern works, but is it only available when you click on the Edit button under the kebab?

Hyperkid123:
@sbuenafe-rh could be. That is your call probably.

Hyperkid123:
@lgalis you are right! i will add the async select thanks.

lgalis:
@Hyperkid123 - thank you, the async filtering is working well - but now there is no option to select 'None' if one wants to remove a workflow from a portfolio item.

Hyperkid123:
:man_facepalming: I will add clear able prop

sbuenafe-rh:
@Hyperkid123 for yah, for the edit icon, we'll have it display when a user clicks on the Edit button under the kebab dropdown. We can probably address this in the Image Uploader jira story. :)

gmcculloug:
Is this PR ready for what it is doing?  Can image upload tasks be moved to a separate PR?

sbuenafe-rh:
@gmcculloug yep image upload tasks can be moved as a separate PR

PR #256, opened 2019-09-10
Fixed post cancel order list and success message

lgalis:
This PR fixes the following error that occurs when cancelling an order:

Screenshot from 2019-09-10 15-45-10


Hyperkid123:
@lgalis works in UI. Can you please fix the tests?

lgalis:
@Hyperkid123 - updated the fix

PR #259, opened 2019-09-18
Fixed add/edit portfolio modal

Hyperkid123:
### Changes
- added clear able option to to workflow select
  - now that the options are loaded asynchronously, server does not return an empty options obviously
  - clear able select has addition button which deletes the value
- Fixed empty description was not saved
  - when value is deleted from formState its value is `undefined` but API expects `null`
- Fixed URL freeze when trying to save portfolio which has `%` in name
  - this was caused by async validation because the API client is not serializing special characters
  - query was `?filter[name]=%%%` instead of `?filter[name]=%25%0A%25%0A%25%0A.`
  - I will create interceptor function in the JS clients library to cover the rest of platform apps

### DIFF
--- a/src/helpers/portfolio/portfolio-helper.js
+++ b/src/helpers/portfolio/portfolio-helper.js
@@ -1,6 +1,6 @@
 import { getAxiosInstance, getPortfolioApi, getPortfolioItemApi } from '../shared/user-login';
 import { CATALOG_API_BASE } from '../../utilities/constants';
-import { PORTFOLIO_ITEM_NULLABLE } from '../../constants/nullable-attributes';
+import { PORTFOLIO_ITEM_NULLABLE, PORTFOLIO_NULLABLE } from '../../constants/nullable-attributes';
 import { udefinedToNull } from '../shared/helpers';
 
 const axiosInstance = getAxiosInstance();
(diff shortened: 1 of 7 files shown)


src/helpers/portfolio/portfolio-helper.js
```diff
@@ -56,7 +56,7 @@ export async function addToPortfolio(portfolioId, items) {
 }
 
 export async function updatePortfolio(portfolioData) {
-  return await portfolioApi.updatePortfolio(portfolioData.id, portfolioData);
+  return await portfolioApi.updatePortfolio(portfolioData.id,  udefinedToNull(portfolioData, PORTFOLIO_NULLABLE));
```
Fryguy: Typo? `udefined`?

src/helpers/shared/user-login.js
```diff
@@ -9,6 +9,11 @@ import { AccessApi, PrincipalApi, GroupApi } from '@redhat-cloud-services/rbac-c
 
 const axiosInstance = axios.create();
 
+const paramSerializer = config => {
+  config.url = config.url.replace(/(?==)*%+/g, value => value.replace(/%/g, '%25%0A'));
```
Fryguy: Why the %0A?   That's a newline character, and should not be needed.

Additionally, I'm concerned this will replace any %encoded values already in there... for example, if `%25` is already in the string, this will make it `%25%0A25` (or without the `0A` part, just `%2525`).  Is there a reason you're not using some off-the-shelf params encoder instead of rolling your own (or was that just for expediency)?

Hyperkid123: `%0A` that is a mistake on my part. Honestly I don't know why how it go there...

This is just a hot fix. I want to add some proper encoding to all values not just `%`. It will be in the JS client package.

PR #260, opened 2019-09-23
Overwrite only catalog apps in spandx.config

rvsia:
**Description**

Because the catalog menu section contains the approval app, which is not a part of the catalog application, spandx needs to be updated to overwrite only catalog apps.

@karelhala


rvsia:
@lgalis @gmcculloug

PR #262, opened 2019-10-01
Adjust edit portfolio payload

Hyperkid123:
### Changes
- extracted read only attributes from patch payloads
- fixed spandx config for serving js/css assets via proxy
- added ellipsis overflow to left side of portfolio item detail 

~~@gmcculloug @syncrou i got a question about required attributes. All attributes in the edit-portfolio-item form are now required by the API + the in portfolio workflow is also now required. Is that intentional?~~

### Before

screenshot-ci foo redhat com_1337-2019 10 07-12_38_06

### After
screenshot-ci foo redhat com_1337-2019 10 07-12_37_02


Hyperkid123:
@lgalis there is something shifty going on. Tests are green locally, so it might take a while until i find what is going on.

Hyperkid123:
@lgalis ok there was an issue with missing endpoint mock :man_facepalming: its green now

PR #265, opened 2019-10-03
Fixed empty page after deleting last portfolio item.

Hyperkid123:
### Changes
- if you delete all items from a page user is redirected to first page

empty-portfolio

### DIFF
--- a/src/smart-components/portfolio/portfolio.js
+++ b/src/smart-components/portfolio/portfolio.js
@@ -70,7 +70,7 @@ const Portfolio = props => {
 
   const handleEmptyPage = () => {
     dispatch({ type: 'setIsFetching', payload: true });
-    return props.fetchPortfolioItemsWithPortfolio(props.match.params.id, defaultSettings)
+    return props.fetchPortfolioItemsWithPortfolio(props.match.params.id, { ...props.pagination, offset: 0 })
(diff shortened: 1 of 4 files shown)


src/smart-components/portfolio/portfolio.js
```diff
@@ -68,9 +68,18 @@ const Portfolio = props => {
     .catch(() => dispatch({ type: 'setCopyInProgress', payload: false }));
   };
 
+  const handleEmptyPage = () => {
+    dispatch({ type: 'setIsFetching', payload: true });
+    return props.fetchPortfolioItemsWithPortfolio(props.match.params.id, defaultSettings)
```
lgalis: We should keep the existing page limit here, and not switch to the default.

Hyperkid123: Right switch to first page but keep the limit :+1:

Hyperkid123:
@lgalis hm I was sure it was working.

But I've also discovered another issue with trying to keep current page. When deleting from multiple pages for instance. I think the correct way would be always fetch first page after deleting portfolio items. I think trying to keep current limit/offset is unfortunately not going to work when you start thinking about the edge cases.

lgalis:
@Hyperkid123 - agree, going to the first page and keeping the page size would be right.

PR #271, opened 2019-10-11
Removed usage of display_name from catalog objects.

Hyperkid123:
UI side of: https://github.com/ManageIQ/catalog-api/pull/428
Depends on: https://github.com/ManageIQ/catalog-api/pull/429

~~Currently i am having an issue with API so I was not able to test it in UI. I want to wait until ill be able to do that.~~

~~(I did contacted Drew about the API issue: `error: "invalid value for "operation", must be one of #{validator.allowable_values}."`)~~

### Changes
- removed usage of `display_name` from catalog entities.

### DIFF
--- a/src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
+++ b/src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
@@ -60,7 +60,7 @@ exports[`<ItemDetailInfoBar /> should render correctly 1`] = `
             className="elipsis-text-overflow"
           >
             <span>
-              baz
+              quux
             </span>
           </div>
         </h6>
@@ -258,7 +258,7 @@ exports[`<ItemDetailInfoBar /> should render correctly with vendor 1`] = `
             className="elipsis-text-overflow"
           >
             <span>
-              baz
+              quux
             </span>
           </div>
         </h6>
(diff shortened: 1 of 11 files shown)


Hyperkid123:
@syncrou i was able to get to the UI and all the displa_name occurrences have been swapped for `name`. Although the API is still sending dsipla_name in payload (dun't know if that is supposed to happen or not)

src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
```diff
@@ -60,7 +60,7 @@ exports[`<ItemDetailInfoBar /> should render correctly 1`] = `
             className="elipsis-text-overflow"
           >
             <span>
-              baz
+              quux
```
syncrou: Is this change tied to the `display_name` removal?

src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
```diff
@@ -258,7 +258,7 @@ exports[`<ItemDetailInfoBar /> should render correctly with vendor 1`] = `
...
```
syncrou: Same

src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
lgalis: yes - the test this snapshot is based on has the portfolio name=quux and display_name=baz - so the snapshot was correctly changed to use the name value

PR #272, opened 2019-10-11
Add products page

Hyperkid123:
Jira ticket: https://projects.engineering.redhat.com/browse/SSP-735

### Changes
- new products page
- can link from products portfolio item to portfolio item detail
- added universal solution to chrome sub menu highlighting

#### TO DO
- [x] use case insensitive filtering once the common api gem is updated
- [x] add link to portfolio item detail page once portfolio id is in portfolio item payload: https://github.com/ManageIQ/catalog-api/pull/432
- [x] add tests

### Preview
products


Hyperkid123:
@gmcculloug got a question about orphaned portfolios. Right now the card title for these is not clickable. My question is. should we display them? Or maybe have a tab for each group?

PR #279, opened 2019-10-17
Add async filtering to portfolios

Hyperkid123:
### Changes
- Added async filtering by name for portfolios
- Portfolios component is now function using hooks
- updated react-router-dom to be able to use router hooks

### Fixes
- fixed default count value for async pagination to 0
  - no longer shows 0 - 50 before the first API request is finished. Now its 0 - 0

### DIFF
--- a/src/smart-components/portfolio/portfolios.js
+++ b/src/smart-components/portfolio/portfolios.js
@@ -17,9 +17,9 @@ import createPortfolioToolbarSchema from '../../toolbar/schemas/portfolios-toolb
 import ContentGalleryEmptyState, { EmptyStatePrimaryAction } from '../../presentational-components/shared/content-gallery-empty-state';
 import asyncFormValidator from '../../utilities/async-form-validator';
 
-const debouncedFilter = asyncFormValidator((value, dispatch, filteringCallback) => {
+const debouncedFilter = asyncFormValidator((value, dispatch, filteringCallback, meta = defaultSettings) => {
   filteringCallback(true);
-  dispatch(fetchPortfolios(value, defaultSettings)).then(() => filteringCallback(false));
+  dispatch(fetchPortfolios(value, meta)).then(() => filteringCallback(false));
 }, 1000);
 
 const portfoliosRoutes = {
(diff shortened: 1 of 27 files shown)


src/smart-components/portfolio/portfolios.js
```diff
@@ -17,102 +15,90 @@ import { fetchPortfolios } from '../../redux/actions/portfolio-actions';
 import PortfolioCard from '../../presentational-components/portfolio/porfolio-card';
 import createPortfolioToolbarSchema from '../../toolbar/schemas/portfolios-toolbar.schema';
 import ContentGalleryEmptyState, { EmptyStatePrimaryAction } from '../../presentational-components/shared/content-gallery-empty-state';
+import asyncFormValidator from '../../utilities/async-form-validator';
+
+const debouncedFilter = asyncFormValidator((value, dispatch, filteringCallback) => {
+  filteringCallback(true);
+  dispatch(fetchPortfolios(value, defaultSettings)).then(() => filteringCallback(false));
```
lgalis: - any filtering will reset the pagination to the default options.
maybe use the pagination option here:
const debouncedFilter = asyncFormValidator((value, dispatch, filteringCallback, pagination) => {
  filteringCallback(true);
  dispatch(fetchPortfolios(value, pagination)).then(() => filteringCallback(false));
}, 1000);


and then:
-------
 const handleFilterItems = value => {
    stateDispatch({ type: 'setFilterValue', payload: value });
    debouncedFilter(value, dispatch, isFiltering => stateDispatch({ type: 'setFilteringFlag', payload: isFiltering }), meta);
  };

gmcculloug: @lgalis Is that the same as the bug I mentioned this morning? 
 https://projects.engineering.redhat.com/browse/SSP-630

Hyperkid123: @lgalis we can only send limit not offset. Offset has to be reset but you are right that limit can be kept.

Hyperkid123: > @lgalis Is that the same as the bug I mentioned this morning?
> https://projects.engineering.redhat.com/browse/SSP-630

It is the same.

Hyperkid123:
@lgalis @gmcculloug limit should be now kept when filtering.

src/smart-components/portfolio/portfolios.js
lgalis: @Hyperkid123 - sounds good.

PR #281, opened 2019-10-21
Removed isRequired from product description

syncrou:
Product description is not required via the `PortfolioItem` model https://github.com/ManageIQ/catalog-api/blob/master/app/models/portfolio_item.rb#L11 so removing the `isRequired` in the schema.

This depends on https://github.com/ManageIQ/catalog-api/pull/441

There is still more work to be done for the edits to fully work.
For any `patch` operation all `readOnly` params need to be stripped from the payload:
```
found unpermitted parameters: :id, :service_offering_source_ref, :service_offering_type, :portfolio_id
```

Product edits will work once https://github.com/ManageIQ/catalog-api/pull/441 is merged and the `readOnly` attributes are removed from the `patch` payload.

### DIFF
--- a/src/forms/edit-portfolio-item-form.schema.js
+++ b/src/forms/edit-portfolio-item-form.schema.js
@@ -11,9 +11,7 @@ const editPortfolioItemSchema = (loadWorkflows) => ({
   }, {
     component: componentTypes.TEXT_FIELD,
     name: 'description',
-    label: 'Description',
-    isRequired: true,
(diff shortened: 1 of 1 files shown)


Hyperkid123:
@syncrou `isRequired` is attribute that changes the component visuals. (that * in label). You also have to remove the validate attribute `validate: [{ type: validatorTypes.REQUIRED }]` from schema.

syncrou:
@Hyperkid123 please re review. I forgot about this one last week.

PR #286, opened 2019-10-31
Sanitize patch requests based on openApi schema

Hyperkid123:
### Changes
- to avoid constant updates of unpermitted when updating catalog entities i added automatic cleanup based on openApi schema from `/api/catalog/v1.0/openapi.json` hosted in cloud services.

Related to: https://projects.engineering.redhat.com/browse/SSP-877 and many others. After this it should be automated.

### DIFF
--- a/src/helpers/shared/helpers.js
+++ b/src/helpers/shared/helpers.js
@@ -1,4 +1,7 @@
 import moment from 'moment';
+import get from 'lodash/get';
+
+import { PORTFOLIO_ITEM_NULLABLE, PORTFOLIO_NULLABLE } from '../../constants/nullable-attributes';
 
 export const scrollToTop = () => document.getElementById('root').scrollTo({
   behavior: 'smooth',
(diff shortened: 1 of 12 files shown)


src/helpers/shared/helpers.js
```diff
@@ -44,3 +47,19 @@ export const udefinedToNull = (entity, keys) => [ ...Object.keys(entity), ...key
...
+
+const nullableMapper = {
+  PortfolioItem: PORTFOLIO_ITEM_NULLABLE,
+  Portfolio: PORTFOLIO_NULLABLE
+};
+
+export const sanitizeValues = (values, entityType, store) => {
+  const schemas = store.getState().openApiReducer.schema.components.schemas;
+  const permittedValues = Object.keys(values)
+  .filter(key => !get(schemas, `${entityType}.properties.${key}.readOnly`))
```
syncrou: Nice!  I know what's going on here.

Hyperkid123:
> @Hyperkid123 - I am not able to navigate between the different Catalog submenus using the Chrome left side navigation.
> Everything else looks good - each individual path works, but not from the left side nav.

This have nothing to do with navigation. Its only called on two endpoints right now. Edit portfolio and edit portfolio item.

I will see what is happening with the navigation.

Hyperkid123:
> @Hyperkid123 - I am not able to navigate between the different Catalog submenus using the Chrome left side navigation.
> Everything else looks good - each individual path works, but not from the left side nav.

Nice catch. I forgot i cannot use component state in event listeners because the function definition is a closure. Should be working OK.

PR #287, opened 2019-11-04
Updated share portfolio modal to look more like mocks.

Hyperkid123:
### Changes
- update share portfolio look

https://projects.engineering.redhat.com/browse/SSP-350
https://marvelapp.com/625jh26/screen/59029719

#### Before
screenshot-ci cloud redhat com-2019 11 04-10_13_15

#### After
screenshot-ci foo redhat com_1337-2019 11 04-10_12_16

### DIFF
--- a/src/presentational-components/shared/loader-placeholders.js
+++ b/src/presentational-components/shared/loader-placeholders.js
@@ -16,7 +16,9 @@ import {
   GalleryItem,
   Form,
   FormGroup,
-  Title
+  TextContent,
+  Text,
+  TextVariants
 } from '@patternfly/react-core';
 
 export const CardLoader = ({ items, ...props }) => (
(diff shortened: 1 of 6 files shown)


src/presentational-components/shared/loader-placeholders.js
```diff
@@ -158,14 +160,19 @@ const FormItemLoader = () => (
...
     <FormGroup fieldId="1">
       <FormItemLoader />
     </FormGroup>
-    <FormGroup fieldId="2">
-      <FormItemLoader />
-    </FormGroup>
-    <Title size="xl">Groups with access</Title>
+    <TextContent>
+      <Text component={ TextVariants.small }>
+        Invite group
```
lgalis: The text here should be 'Groups with access

Hyperkid123:
@lgalis so i wanted to make the groups select async select. But the rbac api does not support all the filtering features we would need to make it work properly. I added the clearable option to the select so the value can be deleted instead of empty option.

lgalis:
@Hyperkid123 - I do not see the clearable option:
Screenshot from 2019-11-05 09-20-04

Hyperkid123:
@lgalis a forgot to push the change :man_facepalming:

PR #289, opened 2019-11-06
[WIP] Refactoring

Hyperkid123:
### Changes
- removed all code which has something to do with workflows
  - since workflow_ref is no longer accessible in catalog entities (https://github.com/ManageIQ/catalog-api/pull/463), it was removed from UI as well 
- removed of styling for react select


Hyperkid123:
Lot of these changes were applied in the meantime

PR #296, opened 2019-11-11
Convert platform templates to use hooks.

lgalis:
Convert the platform template page to use hooks.  Update filtering and specs.

https://projects.engineering.redhat.com/browse/SSP-907

### DIFF
--- a/src/smart-components/platform/platform-templates.js
+++ b/src/smart-components/platform/platform-templates.js
@@ -1,117 +1,147 @@
-import React, { Component, Fragment } from 'react';
+import React, { useEffect, Fragment, useReducer } from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
-import { bindActionCreators } from 'redux';
+import { useDispatch, useSelector } from 'react-redux';
+import { Route, Switch, useParams } from 'react-router-dom';
+import { Section } from '@redhat-cloud-services/frontend-components';
 import debouncePromise from 'awesome-debounce-promise';
 import ToolbarRenderer from '../../toolbar/toolbar-renderer';
 import ContentGallery from '../content-gallery/content-gallery';
-import { scrollToTop, filterServiceOffering } from '../../helpers/shared/helpers';
+import { scrollToTop } from '../../helpers/shared/helpers';
 import PlatformItem from '../../presentational-components/platform/platform-item';
 import { createPlatformsTopToolbarSchema, createPlatformsFilterToolbarSchema } from '../../toolbar/schemas/platforms-toolbar.schema';
 import { defaultSettings, getCurrentPage, getNewPage } from '../../helpers/shared/pagination';
-import { fetchSelectedPlatform, fetchPlatformItems } from '../../redux/actions/platform-actions';
+import ContentGaleryEmptyState from '../../presentational-components/shared/content-gallery-empty-state';
+import {
+  fetchSelectedPlatform,
+  fetchPlatformItems
+} from '../../redux/actions/platform-actions';
+import asyncFormValidator from '../../utilities/async-form-validator';
+import { SearchIcon } from '@patternfly/react-icons/dist/js/index';
 
-class PlatformTemplates extends Component {
-  state = {
-    filterValue: ''
+const initialState = {
+  filterValue: '',
+  isOpen: false,
+  isFetching: true,
+  isFiltering: false
+};
+
+const platformItemsState = (state, action) => {
+  switch (action.type) {
+    case 'setFetching':
+      return { ...state, isFetching: action.payload };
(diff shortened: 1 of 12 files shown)


src/smart-components/platform/platform-templates.js
```diff
@@ -1,133 +1,163 @@
...
-      items: this.props.platformItems
-      .filter(item => filterServiceOffering(item, this.state.filterValue))
-      .map(data => <PlatformItem key={ data.id } { ...data } />),
-      isLoading: this.props.isPlatformDataLoading
+      items: data ? data.map(item => <PlatformItem key={ item.id } { ...item } />) : [],
+      isLoading: isFetching || isFiltering
     };
 
-    let title = this.props.platform ? this.props.platform.name : '';
+    let title = platform ? platform.name : '';
```
Hyperkid123: this should be `const`

We should add linter rule for this.

src/smart-components/platform/platform-templates.js
```diff
@@ -1,133 +1,163 @@
...
+            numberOfItems: paginationCurrent.count,
+            onPerPageSelect: handleOnPerPageSelect,
+            page: getCurrentPage(paginationCurrent.limit, paginationCurrent.offset),
+            onSetPage: handleSetPage,
             direction: 'down'
           }
         }) }/>
-        <ContentGallery { ...filteredItems }/>
+        <ContentGallery title={ title }
+          isLoading={ isFetching || isFiltering }
```
Hyperkid123: This is a duplicate props. It is also present in the filtered items.

I think it should be removed from `filteredItems` and assign both props explicitly, not by using the spread operator.

src/smart-components/platform/platform-templates.js
```diff
@@ -1,133 +1,163 @@
...
-    platform: selectedPlatform,
-    platformItems: platformItemsData && platformItemsData.data,
-    isPlatformDataLoading
-  };
+    );};
+
+  return (
+    <Switch>
+      <Route path={ '/platforms/detail/:id/platform-templates' }
+        render={ renderItems } />
```
Hyperkid123: unnecessary new line

src/smart-components/platform/platform-templates.js
```diff
@@ -1,133 +1,163 @@
...
-    paginationLinks: platformItemsData && platformItemsData.links,
-    paginationCurrent: platformItemsData && platformItemsData.meta,
-    platform: selectedPlatform,
-    platformItems: platformItemsData && platformItemsData.data,
-    isPlatformDataLoading
-  };
+    );};
+
+  return (
+    <Switch>
```
Hyperkid123: If there is only one child route, there is no need for `Switch`

src/smart-components/platform/platform-templates.js
```diff
@@ -1,133 +1,163 @@
...
 };
 
-const mapDispatchToProps = dispatch => bindActionCreators({
-  fetchSelectedPlatform,
-  fetchPlatformItems
-}, dispatch);
-
 PlatformTemplates.propTypes = {
   filteredItems: PropTypes.object,
   isPlatformDataLoading: PropTypes.bool,
```
Hyperkid123: This props is no loner used

src/smart-components/platform/platform-templates.js
```diff
@@ -1,133 +1,163 @@
...
-    count: PropTypes.number.isRequired
+    count: PropTypes.number
   })
 };
 
 PlatformTemplates.defaultProps = {
   platformItems: [],
   paginationCurrent: {
-    limit: 50
+    limit: 50,
```
Hyperkid123: Prop types which came from `mapStateToProps`  are no longer useful.

src/smart-components/platform/platform.js
```diff
@@ -1,48 +1,31 @@
...
   return (
     <Fragment>
       <Switch>
-        <Route path={ `/platforms/detail/:id/platform-templates` } component={ PlatformTemplates }/>
-        <Route path={ `/platforms/detail/:id/platform-inventories` } component={ PlatformInventories }/>
-        <Route render={ () => <Redirect to={ `/platforms/detail/${props.match.params.id}/platform-templates` } /> } />
+        <Route path={ `/platforms/detail/:id/platform-templates` }>
+          <PlatformTemplates/>
+        </Route>
+        <Route path={ `/platforms/detail/:id/platform-inventories` }>
```
Hyperkid123: Can we use the `component`/`render` syntax? So it matches the rest of the application.

lgalis: @Hyperkid123  - I think we should start using this syntax, to 'stay ahead of the curve' - https://reacttraining.com/blog/react-router-v5-1/...

PR #297, opened 2019-11-11
Add the 'Edit approval workflow' to the portfolio detail kebab menu

lgalis:
Add the 'Edit approval' to the portfolio detail kebab menu on the Portfolio detail page, and drop the 'workflow from all menus

Screenshot from 2019-11-12 09-05-32

Screenshot from 2019-11-12 09-05-40

Screenshot from 2019-11-12 09-05-46

### DIFF
--- a/src/presentational-components/portfolio/porfolio-card.js
+++ b/src/presentational-components/portfolio/porfolio-card.js
@@ -38,7 +38,7 @@ const createToolbarActions = (portfolioId, isOpen, setOpen) => [
       </DropdownItem>,
       <DropdownSeparator key="workflow-portfolio-separator"/>,
       <DropdownItem key="workflow-portfolio-action" component={ Link } to={ `/portfolios/edit-workflow/${portfolioId}` }>
-          Edit approval workflow
+          Edit approval
       </DropdownItem>,
       <DropdownSeparator key="share-portfolio-separator"/>,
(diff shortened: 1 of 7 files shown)


gmcculloug:
Please include a screenshot.  Thanks.

gmcculloug:
I would not expect "Edit approval workflow" to be at the bottom of the list.  It should follow the same order as the Portfolio list view:
image

Also, instead of just `Edit` I would suggest qualifying it like we do for Policies under and RBAC group:
image

lgalis:
@gmcculloug - updated screenshot

Hyperkid123:
@gmcculloug correct me if i am wrong. Do you want to also change the wording of the drop down actions?

lgalis:
@Hyperkid123, @gmcculloug - updated screenshots
As discussed, dropped the 'workflow' from the menus and the 'Edit' for teh portfolio remains as is.

Hyperkid123:
@lgalis looking at the images i think that the select styling is a bit off. I will check it out (not relevant for this PR).

Hyperkid123:
So i ran this locally and i see a couple of prop types warnings coming from PF dropdown
```
Failed prop type: Invalid prop `component` supplied to `InternalDropdownItem`.
```

Let me see what is going on there

Hyperkid123:
@lgalis ok both the prop type error and the strange styling around the separator comes from PF

PR #315, opened 2019-11-20
Fixed icon source fallback priority

Hyperkid123:
### Changes
- icon fallback had higher priority than the actual source and was used even if the icon existed


src/presentational-components/shared/card-icon.js
```diff
@@ -37,6 +37,7 @@ const CardIcon = ({ src, height, platformId, sourceTypeId  }) => {
   const [ isLoaded, setLoaded ] = useState(false);
   const [ isUnknown, setUnknown ] = useState(false);
   const  platformList = useSelector(state => (state.platformReducer ? state.platformReducer.platforms : {}));
+  const fallback = sourceTypeId ? platformTypeIcon[sourceTypeId] : defaultPlatformIcon(platformId, platformList);
```
lgalis: I would use a different name here - maybe 'default' or default-icon?

PR #316, opened 2019-11-20
Added ids to portfolio item detail paragraphs.

Hyperkid123:
QE requested some identifiers on these elements.

Ids match the ids from edit form.


gmcculloug:
cc @brahmanim

PR #320, opened 2019-11-21
Orders async filtering

Hyperkid123:
### Changes
- added async filtering to orders view

As you can see the poor performance really starts to show. User is just mostly looking at loading screen. I think that we should bring back the discussions about better way to get the data from server. A basically have to create 3 sequential requests because each sub collection depends on the previous one

orders-async-filtering


Hyperkid123:
@lgalis done

PR #321, opened 2019-11-21
Update readme.md to use RedHatInsights org

gmcculloug:
There are other references that need to be changed, but doing this separate as it does not effect dependencies and restores the travis link.


gmcculloug:
There are other references that need to be changed, but doing this separate as it does not effect dependencies and restores the travis link.

PR #324, opened 2019-11-25
Updated node dependencies

Hyperkid123:
Its about time we update the webpack configuration we use. There are some packages (like `webpack-serve`) which ale no longer supported and we should migrate to supported alternatives.

Also the webpack config is split accross 8 different files. That is not necessary. There is also a lot of configuration that is not being used at all and should be deleted.


config/spandx.config.js
```diff
@@ -1,12 +1,12 @@
-/*global module, process*/
+/*global process*/
```
lgalis: As we update this file, I think we should follow the same pattern that the template frontend starter api does and move this file to a new directory, profiles/local-frontend, like here: https://github.com/RedHatInsights/insights-frontend-starter-app/blob/master/profiles/local-frontend.js - just to keep this consistent across apps.

Hyperkid123: I will take a look after the tests are working properly. :+1:

Hyperkid123:
@lgalis now includes changes from https://github.com/RedHatInsights/catalog-ui/pull/300

lgalis:
@Hyperkid - that was by mistake- will fix

Hyperkid123:
@lgalis thank you for the help.

lgalis:
The platform is down - will merge after it is back up and can run a test.

PR #325, opened 2019-11-25
[WIP]Add support for multiple approval workflows

lgalis:
First draft for this - layout will change.
Screenshot from 2020-01-24 15-51-18

### DIFF
--- a/src/forms/edit-workflow_form.schema.js
+++ b/src/forms/edit-workflow_form.schema.js
@@ -6,7 +6,7 @@ const editWorkflowSchema = (loadWorkflows) => ({
     {
       component: componentTypes.SELECT,
       name: 'workflow',
-      label: 'Approval workflow',
+      label: 'Select workflow',
       loadOptions: asyncFormValidator(loadWorkflows),
       isSearchable: true,
       isClearable: true
--- a/src/helpers/approval/approval-helper.js
+++ b/src/helpers/approval/approval-helper.js
@@ -11,9 +11,7 @@ export const loadWorkflowOptions = (filterValue = '') =>
         filterValue.length > 0 ? `/?filter[name][contains]=${filterValue}` : ''
       }`
     )
-    .then(({ data }) =>
-      data.map(({ id, name }) => ({ label: name, value: id }))
-    );
+    .then(({ data }) => data.map((wf) => ({ label: wf.name, value: wf })));
 
 export const linkWorkflow = (id, resourceObject) =>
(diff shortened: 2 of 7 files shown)


Hyperkid123:
@lgalis there seems to be an issue with the add workflow button if no workflow was selected:
workflow-crash

lgalis:
@Hyperkid123 - thanks, I'll disable the add button when there's no selection.

Hyperkid123:
@lgalis also I found that it is possible to add same workflow multiple times
workflow-crash

Hyperkid123:
Also, I think using multi-select would be enough.

Do we have some mocks for this?

screenshot-ci foo redhat com_1337-2020 01 28-16_22_30

sbuenafe-rh:
@Hyperkid123 there are no mocks for the redesign of this. it's on my backlog. but you are correct by using the multi-select component. 

I'm curious as to what the "Add workflow" primary button does. We should only have one primary button on this modal.

Approval workflow also has been renamed "Approval process" to avoid the confusion with Ansible's job/workflow template terms.

lgalis:
@Hyperkid123 - I have considered the multiselect, but decided to model this based on the role selection in the rbac ui - I had a mock for that. I do not look of the multiselect. The workflows do have a description - which I would like to add to the table ( it is not returned for the assigned workflow now, but I can look it up).

lgalis:
@sbuenafe - @Hyperkid123 screenshot is not what is in this PR - the screenshot in the description is.
If you want to go with the multiselect, then I will change to that.

lgalis:
@sbuenafe-rh - @Hyperkid123's screenshot is not what is in this PR - the screenshot in the description is.
If you want to go with the multiselect, then I will change to that - there will be no 'Add workflow' button if we do that.
Also the 'Approval process' change - was that the final decision?

sbuenafe-rh:
@lgalis "Approval process" term was vetted with John and with UXD content strategist. If the term changes again, then we can revisit it. Using process instead of workflow, we can bypass the confusion with Ansible's use of "workflow".

gmcculloug:
@sbuenafe-rh The term "Approval process" term impacts the Approval UI as well since we use the term "Workflow" in several places there.  I can open an issue for this, but do we have mocks or the term to use there?  If you are already in the Approval UI would you refer to them as "Processes"?

lgalis:
Closing this PR - Replaced by https://github.com/RedHatInsights/catalog-ui/pull/390

PR #335, opened 2019-12-09
Add async select to the share portfolio modal

lgalis:
Add async select to the share portfolio modal

Screenshot from 2019-12-09 14-58-57

### DIFF
--- a/src/forms/form-fields/share-group-select.js
+++ b/src/forms/form-fields/share-group-select.js
@@ -8,7 +8,6 @@ export const ShareGroupSelect = ({
   FieldProvider,
   inputName,
   selectName,
-  groups,
   loadOptions,
   permissions
 }) => {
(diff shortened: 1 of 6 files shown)


Hyperkid123:
@lgalis btw i might have possible fix for the overflowing context menu.

src/forms/form-fields/share-group-select.js
```diff
@@ -34,6 +38,7 @@ ShareGroupSelect.propTypes = {
   inputName: PropTypes.string.isRequired,
   selectName: PropTypes.string.isRequired,
   groups: PropTypes.any,
```
Hyperkid123: This prop is no longer necessary

src/forms/form-fields/share-group-select.js
```diff
@@ -34,6 +38,7 @@ ShareGroupSelect.propTypes = {
...
+  loadOptions: PropTypes.func,
```
Hyperkid123: This should be required

lgalis:
@Hyperkid123 - feedback addressed

PR #338, opened 2019-12-12
Added basic survey editing.

Hyperkid123:
### Changes
- added survey editor to portfolio items
- editing works

### What is missing
- ~~turn off adding and changing order of fields (will be done in DDF)~~ done
- turn of removing fields
- validation editing (Will be done in DFF)
- visual design (Susan is working on that)

survey editing

### DIFF
--- a/src/helpers/shared/user-login.js
+++ b/src/helpers/shared/user-login.js
@@ -10,7 +10,8 @@ import {
   PortfolioItemApi,
   OrderApi,
   OrderItemApi,
-  IconApi
+  IconApi,
+  ServicePlansApi
 } from '@redhat-cloud-services/catalog-client';
 
 import {
@@ -77,6 +78,11 @@ const topologicalInventoryApi = new TopologicalDefaultApi(
   axiosInstance
 );
 const iconApi = new IconApi(undefined, CATALOG_API_BASE, axiosInstance);
+const servicePlansApi = new ServicePlansApi(
+  undefined,
+  CATALOG_API_BASE,
+  axiosInstance
+);
 
 export function getSourcesApi() {
   return sourcesApi;
@@ -136,6 +142,10 @@ export function getAxiosInstance() {
 
 export function getIconApi() {
   return iconApi;
+}
+
+export function getServicePlainsApi() {
+  return servicePlansApi;
 }
 
 const grapqlInstance = axios.create();
--- a/src/helpers/shared/user-login.js
+++ b/src/helpers/shared/user-login.js
@@ -144,7 +144,7 @@ export function getIconApi() {
   return iconApi;
 }
 
-export function getServicePlainsApi() {
+export function getServicePlansApi() {
   return servicePlansApi;
 }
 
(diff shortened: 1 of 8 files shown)


Hyperkid123:
@gmcculloug since the Visuals are obviously not complete and it will take some time to do design them, should we hide the link and leave the route there? I would do that for now.

gmcculloug:
Good start, excited to see where this goes.

src/smart-components/portfolio/portfolio-item-detail/detail-toolbar-actions.js
```diff
@@ -93,6 +94,15 @@ const DetailToolbarActions = ({
                   </Link>
                 }
                 role="link"
+              />,
+              <DropdownItem
+                aria-label="Rdit survey"
```
lgalis: Typo - 'Rdit'

src/helpers/shared/user-login.js
```diff
@@ -138,6 +144,10 @@ export function getIconApi() {
   return iconApi;
 }
 
+export function getServicePlainsApi() {
```
lgalis: typo - getServicePlansApi, not getServicePlainsApi

src/smart-components/survey-editing/survey-editor.js
```diff
@@ -0,0 +1,149 @@
...
+import FormBuilder from '@data-driven-forms/form-builder';
+import {
+  builderMapper,
+  fieldProperties,
+  pickerMapper,
+  propertiesMapper
+} from '@data-driven-forms/form-builder/dist/pf4-builder-mappers';
+import {
+  getAxiosInstance,
+  getServicePlainsApi
```
lgalis: type, as above - getServicePlansApi

src/smart-components/survey-editing/survey-editor.js
```diff
@@ -0,0 +1,149 @@
...
+  const [editedTemplate, setEditedTemplate] = useState({ fields: [] });
+  useEffect(() => {
+    getAxiosInstance()
+      .get(
+        `${CATALOG_API_BASE}/portfolio_items/${portfolioItemId}/service_plans`
+      )
+      .then(([{ create_json_schema: { schema } }]) => setSchema(schema));
+  }, []);
+  const handleSaveSurvey = () => {
+    return getServicePlainsApi()
```
lgalis: - same - getServicePlansApi

Hyperkid123:
@lgalis yep this will all happen later. I waiting with that for Susan. We dont event know if we will use normal page of some overlay etc.

Multi select is not supported yet

> I do not see the component chooser at all when opening the editor.

That is intentional. You are not supposed to be able to add/change/remove fields. Removing is still visible though and will change in the future.

lgalis:
@Hyperkid123  - sounds good - I mentioned the component chooser because it is visible in your gif - so that is different than what I saw when testing.

gmcculloug:
cc @sbuenafe-rh

Hyperkid123:
@lgalis i got fix fir the multi select just updating to new package

Hyperkid123:
OK the multi select should be crashing anymore

PR #345, opened 2019-12-13
Disabled non-existing products links.

Hyperkid123:
### Changes
- links in products pages are now disabled
  - links are disabled if source or portfolio of portfolio item no longer exist
- fixed links for existing items (it was using the old non query params links)

### Preview

screenshot-ci foo redhat com_1337-2019 12 13-11_17_00


gmcculloug:
I think this change is good, but wanted to let you know that soon these non-existent Products will not be returned from the api, but first we need to complete the work to move the Portfolio ACLs out of RBAC.  Then the api will properly filter products based on the portfolios a user has access to.

PR #371, opened 2020-01-13
Empty states when server has no data

Hyperkid123:
### Changes
- added redux middleware which adds `noData` flag
  - the flag is set to true when API response count is 0 and no filter was applied
- added empty states variants for `noData` scenario
- completely removed toolbar actions when `noData` is set. User will use primary actions from empty states

The same pattern is applied to all entities with exception of platforms (There are some missing API features to enable pagination/filter)

### Portfolio with items
screenshot-ci foo redhat com_1337-2020 01 13-13_43_33

### Portfolio with no items
screenshot-ci foo redhat com_1337-2020 01 13-13_42_49

### DIFF
--- a/src/helpers/shared/pagination.js
+++ b/src/helpers/shared/pagination.js
@@ -1,7 +1,8 @@
 export const defaultSettings = {
   limit: 50,
   offset: 0,
-  count: 0
+  count: 0,
+  filter: ''
 };
 
 export const getCurrentPage = (limit = 1, offset = 0) =>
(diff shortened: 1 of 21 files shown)


lgalis:
@Hyperkid123 - this looks great - just a question regarding the icon/wording for the empty state - it is different than the mock ( https://marvelapp.com/625jh26/screen/58832623 ) - are the mocks not up to date for the empty state?

Hyperkid123:
@lgalis I will check that out.

lgalis:
@Hyperkid123  - I will merge this as is - you can replace the icon and wording in a separate PR

PR #372, opened 2020-01-13
Platform service offering

Hyperkid123:
### Changes
- added detail page for source service offering
  - this was brought up during the fit-n-finish
  - no design for it just yet, but it should be pretty similar to the portfolio item detail

sorvice-offering


gmcculloug:
Looks good.  👍

Hyperkid123:
I just put all the "useful" data there are so it won't look so empty.

PR #376, opened 2020-01-15
Added description to portfolio detail toolbar.

Hyperkid123:
I also moved the styles to 'App.scss' and removed it from App.js. The platform is adding the style sheet to the document.

### Before
screenshot-ci cloud redhat com-2020 01 15-09_16_47

### After
screenshot-ci foo redhat com_1337-2020 01 15-09_16_13

cc @sbuenafe-rh


sbuenafe-rh:
Portfolio description styling should be body text. 

Reference: https://www.patternfly.org/v4/documentation/core/demos/cardview/basic

PR #380, opened 2020-01-15
Changed copy portfolio select to async select with filter.

Hyperkid123:
### Changes
- use async select for portfolio list in copy portfolio modal

copy-portfolio-async


sbuenafe-rh:
Looks good, but the Cancel button should have a link text styling.

Hyperkid123:
@sbuenafe-rh fixed

screenshot-ci foo redhat com_1337-2020 01 15-17_01_05

PR #390, opened 2020-01-28
Set multiple approval processes (workflows) for resources

lgalis:
Set multiple approval processes (workflows) for resources:

Screenshot from 2020-01-28 16-11-51

Notifications combined into one:
Screenshot from 2020-01-29 14-51-43

### DIFF
--- a/src/smart-components/common/edit-approval-workflow.js
+++ b/src/smart-components/common/edit-approval-workflow.js
@@ -57,9 +57,6 @@ const EditApprovalWorkflow = ({
   }, []);
 
   const onSubmit = (formData) => {
-    console.log( 'Debug - data, formData', data, formData);
-    console.log( 'Debug - id, objectId', id, objectId);
-
     history.push(pushParam);
     const toUnlinkWorkflows = data.filter(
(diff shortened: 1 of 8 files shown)


Hyperkid123:
We are getting one notification per workflow action right now:
multiple-notifications

Maybe it could be switched for something like 5 workflows were linked and 1 unlinked or something similar. Or just have One notification. Right now you can basically fill the whole screen with notifications.

src/smart-components/common/edit-approval-workflow.js
```diff
@@ -56,43 +56,52 @@ const EditApprovalWorkflow = ({
     ).then(() => stateDispatch({ type: 'setFetching', payload: false }));
   }, []);
 
-  const onSubmit = (values) => {
+  const onSubmit = (formData) => {
     history.push(pushParam);
-    const approvalWorkflow = data ? data[0] : undefined;
+    const toUnlinkWorkflows = data.filter(
```
Hyperkid123: I think it might be easier to use initialValue of the field from the second argument you get to the onSubmit function, to determine the linked/unlinked workflows.
```jsx
const onSubmit = (formData, formApi) => {
  formApi.getState().initialValues.fieldName
  ...
```

PR #396, opened 2020-01-30
Added form state to survey editing and restoreable options.

Hyperkid123:
### Changes
- form state handling to the form preview (not validations right now)
- restore select options from base schema

form-state-survey-editor

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2064,9 +2064,9 @@
       }
     },
     "@babel/runtime-corejs2": {
-      "version": "7.8.3",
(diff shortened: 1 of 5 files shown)


gmcculloug:
Love this 😍.  thanks @Hyperkid123

Hyperkid123:
Have to see what is up with the tests :thinking:

Hyperkid123:
Ah I know why it is failing :D I am simulation mouse click instead of form submit. We updated that in data-driven forms so you can use Enter key to submit forms...

PR #404, opened 2020-02-05
Bundle size optimizations

Hyperkid123:
~~**This pr currently uses beta releases of frontend-components packages to test if it works and serve as an example to other teams.**~~

### Changes
- use babel transform import plugins to change import paths to absolute paths
  - should be eventually done manually
- update necessary packages which implement build changes
- react-tokens are still an issue. Pr was created and merged and is available in react-core version `3.136.3` which is a pre-release
- set up code splitting to reduce time to first render
  - the largest chunk is now under 1MB and I will work in future PR to reduce it further.

### Chunk distributions and size
screenshot-127 0 0 1_8888-2020 02 11-09_06_59


Hyperkid123:
cc @rvsia @karelhala

PR #418, opened 2020-02-12
Fixed overflow in share portfolio modal.

Hyperkid123:
### Changes
- ~~added max with to the select value container~~
- use grid layout to enforce correct width
- enabled portals for select menus to enable menu overflow over the modal container
- use render function instead of children in FieldProvider to prevent re-initialization after select change

### Before
share modal select sizing

### After
screenshot-ci foo redhat com_1337-2020 02 12-16_46_14

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -325,26 +325,6 @@ a.pf-c-breadcrumb__item {
   height: 16px;
 }
 
-.share-group-modal{
-  .height-normalize {
-    height: 38px;
-  }
-  .share-column {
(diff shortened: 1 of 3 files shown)


lgalis:
@Hyperkid123  - I still see this :

image

Hyperkid123:
@lgalis I have changed the layout to enforce the fixed-width ratio between these columns.
screenshot-ci foo redhat com_1337-2020 02 19-08_39_28

PR #425, opened 2020-02-17
Migrate to styled components

Hyperkid123:
Since we no longer build any CSS files and are injected o the DOM we can take it a step further to save some resources and improve performance.

Before these changes, all CSS were injected at once to the DOM even though some rules were never used. With styled-components, each component has its own CSS and it is injected only when the component first use.  Also, it eliminates the possibility of dead CSS rules, because the styles are the actual components (i have found dozens of dead CSS rules in our style sheet).

Some styles could not be replaced with styled-components, but those were added to the global styled sheet.


lgalis:
@Hyperkid123 - please rebase and I will merge the PR

PR #435, opened 2020-02-25
Add the 'delete' permission verb to the order/update sharing option

lgalis:
Add the 'delete' verb to the order/update sharing option.
No UI change for the user - but selecting the 'order/update' option for sharing will now include the 'delete' verb: "delete,order,read,update'


gmcculloug:
Is there a reason we would not want to show "Delete" in the UI as well?

lgalis:
@gmcculloug - the order/edit before this PR had these verbs: 'order,read,update'
 - I believe the 2 options are there for ease of use

1.We could argue that 'edit' implies delete as well -in which case we leave the 2 options
or 
2.We could have 3 options:
order/view => 'order, read'
order/edit => 'order,read,update'
??? order/edit/delete => 'order, read,update,delete'

/cc @sbuenafe-rh

Hyperkid123:
> Is there a reason we would not want to show "Delete" in the UI as well?

If we would go forward with this option, I would strongly recommend using multi-select instead of single select with the view/edit/... labels.

lgalis:
The create portfolio, copy portfolio and delete portfolio are only allowed for Catalog Admins. 
The delete permission is not shareable. Closing this PR.

PR #436, opened 2020-02-26
Add portfolio name to product

Hyperkid123:
### Changes
- add portfolio name to the product card

### Possible issues
What about portfolio items without any portfolio? Currently, it does not display the portfolio name and there is also no link since we don't have enough data to show the portfolio item detail. Should we filter out orphaned portfolio items?

portfolio-name-product

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2195,7 +2195,7 @@
         "react-select": {
           "version": "3.0.8",
           "resolved": "https://registry.npmjs.org/react-select/-/react-select-3.0.8.tgz",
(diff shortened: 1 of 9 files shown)


gmcculloug:
> What about portfolio items without any portfolio?
The APIs have been updated so you can no longer create a portfolio item without adding it to a portfolio.  For a long time you could, but we did not have a use case and it just ends up causing problems if we allowing it.

I have removed some of these orphaned portfolio items in the past and will take another pass on CI and QA environments this morning to fully remove them.

PR #441, opened 2020-02-27
Remove order button when source is unavailable.

Hyperkid123:
https://projects.engineering.redhat.com/browse/SSP-1283

I went with alert instead of disabling the button. The user must know why the button is available and popovers are totally buggy in PF. But of course, it can be changed if you decide the disabled button is preferable.

Also, I fixed a typo that was forcing the portfolio item name to be aligned to the bottom.

### Available source:
screenshot-ci foo redhat com_1337-2020 02 27-09_48_34

### Unavailable source:
screenshot-ci foo redhat com_1337-2020 02 27-09_40_24

Screenshot from testing:
Screenshot from 2020-02-28 12-25-46

### DIFF
--- a/src/presentational-components/styled-components/level.js
+++ b/src/presentational-components/styled-components/level.js
@@ -2,5 +2,5 @@ import styled from 'styled-components';
 import { LevelItem } from '@patternfly/react-core';
 
 export const StyledLevelItem = styled(LevelItem)`
-  align-items: ${({ alingEnd }) => (alingEnd ? 'end !important' : 'inherit')};
+  align-items: ${({ alignEnd }) => (alignEnd ? 'end !important' : 'inherit')};
 `;
--- a/src/smart-components/portfolio/portfolio-item-detail/detail-toolbar-actions.js
+++ b/src/smart-components/portfolio/portfolio-item-detail/detail-toolbar-actions.js
@@ -1,6 +1,7 @@
 import PropTypes from 'prop-types';
 import React, { Fragment } from 'react';
 import {
+  Alert,
   Dropdown,
   DropdownItem,
   DropdownPosition,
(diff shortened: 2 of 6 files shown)


gmcculloug:
I'm ok with the alert if @sbuenafe-rh is.  I would like to modify the message as the current "Source for this product is no longer available" makes it sound like it was removed.

A few thoughts:
1. "Source" should be "Platform" when referenced from Catalog.  (Need to review this going forward as it gets confusing referring to the same resource different ways depending on context.)

2. If the API returns `availability_status = "unavailable"` a better message would be: 
"Platform for this product is unavailable"

3. If the Source is actually removed and no longer returned by the API it would be great if the alert said:
"Platform for this product is no longer available"

Hyperkid123:
@gmcculloug agree with 1. and 2. 

Regarding the 3. How does a response look like if there is no Source? Can we even query sources like that or do we get 404? Or is it a source with the archived value?

gmcculloug:
I did a quick test and it looks like you will get a 404 if the sources has been deleted.

Hyperkid123:
Ok so for that I would have to slightly change the logic of the screen data loading. Currently, it will throw you out because the source does not exist.

gmcculloug:
Understood, that was not the target of this change so we can push that work into a separate story and PR if you want.

Hyperkid123:
I also talked to @sbuenafe-rh and she will check the wording with one of her colleagues

sbuenafe-rh:
@Hyperkid123 I would make the following changes. 

If a source is down, then I would disable the Order button. (I wouldn't say **Offline**, like it does in the mocks). I would also place an inline alert between the Header title div (Icon + Header title) and the main body. 

I'm still working on the wording.

gmcculloug:
@Hyperkid123 @sbuenafe-rh Any issue with getting this cleaned up and merge and handle the updated messages as a followup PR?

Hyperkid123:
@gmcculloug VPN was down co I could not take screenshots but the work that @sbuenafe-rh suggested is done. Let me check again.

gmcculloug:
Great, if @lgalis can review I'd like to get this in as I think it is a really nice user experience improvement.  Still need a followup API check in place for direct API callers, but focus is on UI for GA.

Hyperkid123:
@gmcculloug sorry I can't log in, so I can't take pictures.

gmcculloug:
np.  @lgalis is this something you can take a look at.  Let me know if you need help setting up a test.  We can update the screenshot post-merge if needed.

lgalis:
@gmcculloug - yes, I checked this branch out yesterday - was waiting for the final word from @sbuenafe-rh . Will run one test and merge

PR #442, opened 2020-03-02
Enable survey reset

Hyperkid123:
### Changes
- added button to modified surveys to sync with topology (removes all changes made by user and uses base survey)
- fixed some console errors tied to incorrect prop types

survey-sync

### DIFF
--- a/src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
+++ b/src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
@@ -216,7 +216,7 @@ exports[`<ItemDetailInfoBar /> should render correctly 1`] = `
                     zIndex={9999}
                   >
                     <span>
-                      11 months ago
+                      12 months ago
                     </span>
                     <Portal
                       containerInfo={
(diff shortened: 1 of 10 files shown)


Hyperkid123:
@gmcculloug @sbuenafe-rh again I have a question about the wording I've used. Synchronize with Ansible. I want to avoid the word "Reset" sync that is something that should be associated with a session reset (get the user back to the same state when they opened the editor).

Is this ok or should we use something else?

gmcculloug:
What do you think about "Restore to Ansible Tower version"?

My concern with "Synchronize" is that users might think that we are preserving their settings while importing newer data from Tower, when in fact this is a hard reset to what is currently in Tower.

Hyperkid123:
changed the title to what @gmcculloug suggested

src/test/smart-components/portfolio/portfolio-item-detail/__snapshots__/item-detail-info-bar.test.js.snap
```diff
@@ -216,7 +216,7 @@ exports[`<ItemDetailInfoBar /> should render correctly 1`] = `
                     zIndex={9999}
                   >
                     <span>
-                      11 months ago
+                      12 months ago
```
lgalis: @Hyperkid123 - the test that checks this snapshot will pass now, but fail next month. I would either remove the test or remove the dates and regenerate the snapshot without the dates

Hyperkid123: I will add a mock to the test that will return always one year old date. That should be enough.

PR #443, opened 2020-03-02
Replace frontend components spinner with PF spinner.

Hyperkid123:
Spinner looks like this: https://www.patternfly.org/v4/documentation/react/components/spinner


lgalis:
@Hyperkid123 - I will merge after the conflicts are resolved.

PR #448, opened 2020-03-05
Add alert if order resource fails to load.

Hyperkid123:
https://projects.engineering.redhat.com/browse/SSP-1311
### Changes
- order detail will no longer redirect the user if order external resource fails to load
- PF alert is displayed with a warning that specific resource is not available

screenshot-ci foo redhat com_1337-2020 03 05-10_54_55


gmcculloug:
👍

Hyperkid123:
@lgalis Yes, I did not realize that this is possible. I will add it.

Hyperkid123:
@lgalis portfolio alerts are now handled as well

PR #449, opened 2020-03-05
Handle source 404 in portfolio item detail.

Hyperkid123:
https://projects.engineering.redhat.com/browse/SSP-1302
### Changes
- when the portfolio item source is no longer available, show a warning alert and disable the order button

screenshot-ci foo redhat com_1337-2020 03 05-16_54_38

### DIFF
--- a/src/smart-components/portfolio/portfolio-item-detail/portfolio-item-detail.js
+++ b/src/smart-components/portfolio/portfolio-item-detail/portfolio-item-detail.js
@@ -51,8 +51,20 @@ const PortfolioItemDetail = () => {
     );
   }
 
-  const uploadIcon = (file) => uploadPortfolioItemIcon(portfolioItem.id, file);
+  const availability = source.availability_status || 'unavailable';
 
+  const unavailable = [source]
+    .filter(({ notFound }) => notFound)
+    .map(({ object }) => (
+      <Alert
+        className="pf-u-mb-sm"
+        key={object}
+        variant="warning"
+        isInline
+        title={`The ${object} for this product is no longer available`}
+      />
+    ));
+  const uploadIcon = (file) => uploadPortfolioItemIcon(portfolioItem.id, file);
   return (
     <Fragment>
       <Switch>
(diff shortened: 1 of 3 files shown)


gmcculloug:
@sbuenafe-rh Are you still looking into getting the text reviewed for this case and when the sources is unavailable?  Any ETA on that?

sbuenafe-rh:
@gmcculloug was following up on it with other apps. If this is the Catalog Admin and/or the Catalog User, would they be able to view Sources to rectify the situation?

We can merge the PR and file a bug to get the text updated.

Hyperkid123:
@sbuenafe-rh this message will appear if the source no longer exists. It was deleted from Sources. I don't know if there is hard or soft (archive) delete.

gmcculloug:
@sbuenafe-rh Today, only org_admins on the platform can make changes to sources.  RBAC permissions do not effect this at all.

src/smart-components/portfolio/portfolio-item-detail/portfolio-item-detail.js
```diff
@@ -77,15 +78,24 @@ const PortfolioItemDetail = () => {
...
-              availability={source.availability_status}
+              availability={availability}
             />
+            {source.notFound && (
+              <Alert
+                className="pf-u-ml-lg pf-u-mr-lg"
+                id="source-unavailable-alert-info"
+                variant="warning"
+                isInline
+                title="Platform for this product is no longer available"
```
lgalis: The title here should match the one in line 98

Hyperkid123: These are actually different cases. This platform is just deleted, but the alert bellow has an existing source but its status is `unavailable`. Susan is working someone on the terminology.

lgalis:
@Hyperkid123 - The last commit handling the 404 for non-existent portfolios - that was for PR 448, not this one, right? For this PR - just adding the 'The' in the front of the title was the only change left.

Hyperkid123:
Yup, I've mixed them up. :sleepy:

PR #454, opened 2020-03-09
Display Create portfolio button based on permissions

lgalis:
Display the Create portfolio button/menu item based on permissions:

Screenshot from 2020-03-10 16-05-50
Screenshot from 2020-03-10 16-05-22
Screenshot from 2020-03-11 23-36-32

### DIFF
--- a/src/helpers/shared/helpers.js
+++ b/src/helpers/shared/helpers.js
@@ -71,3 +71,10 @@ export const readableBytes = (bytes) => {
 
   return (bytes / Math.pow(1024, i)).toFixed(2) * 1 + ' ' + sizes[i];
 };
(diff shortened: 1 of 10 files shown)


src/helpers/shared/helpers.js
```diff
@@ -71,3 +71,10 @@ export const readableBytes = (bytes) => {
 
   return (bytes / Math.pow(1024, i)).toFixed(2) * 1 + ' ' + sizes[i];
 };
+
+export const hasPermission = (userPermissions, permissions) =>
```
Hyperkid123: this function is already in catalog route. Can you extract it and use it?

PR #459, opened 2020-03-11
Add autorefresh for approval requests in order detail screen.

Hyperkid123:
### Changes
If there are no approval requests for order item, this spinner will show and it will call the API periodically every 3sec, until there are some.

## We should not use this a patter. This is just a placeholder until there is a better way to listen to async tasks from UI.

approval-polling


gmcculloug:
"Approval is 🐝 ing setup"?  😉 

How about "Creating approval request"?

PR #463, opened 2020-03-17
Prefetch select initialValues on first render to avoid missing options.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-924

### Issues

Select expects its initial values to be included in the first `loadOptions` call. That is required due to incomplete information it gets. Initial values are just literals but options are objects with value and label. So if the initial value is not in options it gets removed. And because API returns only first 100 items, values from second pages were removed on the first render.

Now, the initial load options request for approval filters specifically for selected values alongside the rest of the options.

I tested it by limiting the workflow options request to one or two records.

### DIFF
--- a/src/helpers/approval/approval-helper.js
+++ b/src/helpers/approval/approval-helper.js
@@ -4,16 +4,20 @@ import { defaultSettings } from '../shared/pagination';
 
 export const getApprovalWorkflows = () => getWorkflowApi().listWorkflows();
 
-export const loadWorkflowOptions = (filterValue = '') =>
-  getAxiosInstance()
+export const loadWorkflowOptions = (filterValue = '', initialLookup = []) => {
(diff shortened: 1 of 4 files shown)


lgalis:
@Hyperkid123  - LGPortfolio has the hsong-process workflow set:

with 5 page limit for workflow options:
Screenshot from 2020-03-17 16-05-38

with 500 page limit for workflow options:
Screenshot from 2020-03-17 16-05-12

Hyperkid123:
Let me double-check

Hyperkid123:
Ok, there was an issue with the debounce and concurrency. Since the function is debounced, only the last promise is actually resolved so the lookup for initial values cannot be done in parallel. Now they are chained and the data is showing correctly. It's a slight performance hit, but only on first options request. After that, it's the same as before.

PR #473, opened 2020-03-19
Fixed missing permissions for add products button.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1362

### Issues
- no permissions on add products button in portfolio empty state

@gmcculloug I've used `catalog:portfolio_items:create` for now. I expect that this will be a part of user capabilities later on. We should probably also prepare some messages for the users if they don't have permission to add portfolio items.


gmcculloug:
@mkanoor See PR description.  I think this makes sense for now.  Do you agree?

PR #475, opened 2020-03-24
Portfolio user capabilities

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1371

### Changes
- switch to API version 1.1
- hide portfolio actions behind portfolio user_capabilities
- hide routes behind portfolio user_capabilities
- update babel to enable optional chaining

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,22 +13,78 @@
         "@babel/highlight": "^7.0.0"
       }
     },
-    "@babel/core": {
-      "version": "7.7.4",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.7.4.tgz",
-      "integrity": "sha512-+bYbx56j4nYBmpsWtnPUsKW3NdnYxbqyfrP2w9wILBuHzdfIKz9prieZK0DFPyIzkjYVUe4QkusGL07r5pXznQ==",
+    "@babel/compat-data": {
+      "version": "7.9.0",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.9.0.tgz",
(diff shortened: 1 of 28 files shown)


Hyperkid123:
@lgalis there are no capabilities for that yet. We only have copy, update, show, destroy and create capabilities available at this moment.

gmcculloug:
@mkanoor correct me if I'm wrong but 'Add products' and 'Remove products' should be based on having update permissions on the portfolio.  We are not planning on exposing specific capabilities for each action at the moment.

Setting approval is still and outstanding item of work.

lgalis:
@gmcculloug  - agree - the Add products/ Remove products are based on the 'update' permissions on the portfolio. 
@Hyperkid123  - are the Add products/Remove products buttons not yet hidden based on this in this PR? will this work be part of the separate PR?

gmcculloug:
Also meant to note that there are some issues with the data coming back in `user_capabilities` section which we are working to address.  So the code might be doing the right thing based on the response.

@lgalis You can check the api response and also test against a user that has limited portfolio permissions.  I'll DM you that info to test with.

Hyperkid123:
> @gmcculloug - agree - the Add products/ Remove products are based on the 'update' permissions on the portfolio.
> @Hyperkid123 - are the Add products/Remove products buttons not yet hidden based on this in this PR? will this work be part of the separate PR?

I will add it to this one. Thanks for the info.

Hyperkid123:
Ok the add/remove products should no longer be displayed for users without the update capability.

Hyperkid123:
> As a follow up - we should remove the checkboxes on the card when the 'update' permission is not available.

That is a part of the portfolio items pr

PR #476, opened 2020-03-24
Portfolio item user capabilities

Hyperkid123:
waiting for: https://github.com/RedHatInsights/catalog-ui/pull/475

jira: https://projects.engineering.redhat.com/browse/SSP-1372, https://projects.engineering.redhat.com/browse/SSP-1376

### Changes
- added restrictions on portfolio items based on user capabilities
- added user capabilities check on catalog route
- fixed up some prop types errors

@lgalis @gmcculloug I used the portfolio item `destroy` capability do hide/show the checkbox for selecting the card. I know that @lgalis mentioned that it should be based on the portfolio `update` capability. Just want to double-check which one of these flags is the correct one,

### DIFF
--- a/src/helpers/shared/helpers.js
+++ b/src/helpers/shared/helpers.js
@@ -72,9 +72,7 @@ export const readableBytes = (bytes) => {
   return (bytes / Math.pow(1024, i)).toFixed(2) * 1 + ' ' + sizes[i];
 };
 
-export const hasPermission = (userPermissions, permissions) =>
-  userPermissions
-    ? permissions.every((permission) =>
-        userPermissions.find((item) => item.permission === permission)
-      )
-    : false;
+export const hasPermission = (userPermissions = [], permissions = []) =>
+  permissions.every((permission) =>
+    userPermissions.find((item) => item.permission === permission)
+  );
--- a/src/redux/reducers/portfolio-reducer.js
+++ b/src/redux/reducers/portfolio-reducer.js
@@ -31,7 +31,11 @@ export const portfoliosInitialState = {
     }
   },
   portfolioItem: {
-    portfolioItem: {}
+    portfolioItem: {
+      metadata: {
+        user_capabilities: {}
+      }
+    }
   },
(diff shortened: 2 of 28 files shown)


lgalis:
@Hyperkid123 - in the portfolio detail page - when a user does not have the 'update' portfolio permission, there is no remove button, so although there are checkboxes on the product cards, there is nothing for the user to do with the selected cards. This is why I said that the checkbox on these cards need to be based on the 'update' permission for the parent portfolio - for this case at least.

Hyperkid123:
@lgalis I get that. I am just wondering if it's possible to have "global" permission to delete portfolio items, but individual products can "overrule" this permission. So there is, for instance, one portfolio item in the set cannot be deleted but the rest can.

gmcculloug:
@Hyperkid123 There is no global permission currently or on the roadmap.  The ability to delete portfolio_items is only controlled by the access defined on the portfolio.

Hyperkid123:
@gmcculloug so the `destroy` flag on portfolio items means something else?

Hyperkid123:
Ok I have swapped to portfolio `update` flag instead of portfolio item `destroy` flag.

Thanks for the clarification.

gmcculloug:
@Hyperkid123 that's a good point and something I forgot about.  Since we have an object (portfolio_item) we can define the "delete" action as part of the objects `user_capabilities` metadata.  For a portfolio_item this is always based on having update access to the portfolio today. 

So your point is well-taken but we do not have a use case today that supports different delete values for all the portfolio_items in a single portfolio.

Does that help any?

Hyperkid123:
I don't think we have any use case like that. I just assumed that the destroy flag was the one that handles the removal from the portfolio. 

Its all clear to me now :+1:

Hyperkid123:
Ok, I have found the issue. It was in the permission calculation if the required user capability was `undefined`.

I can now access the portfolio item detail with catalog users. Nice catch @lgalis.

PR #477, opened 2020-03-24
Update year in test to  fix snapshot

lgalis:


syncrou:
Is there a way for the date year to be dynamic? I’m assuming it failed because that date wasn’t a Friday in 2018?

lgalis:
@syncrou - the chnage is to set the current date to the one in the test so the relative date is always 'justNow'

PR #480, opened 2020-03-25
Add different platform empty state message for non org admin.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1377

screenshot-ci foo redhat com_1337-2020 03 25-14_27_51


gmcculloug:
cc @sbuenafe-rh

PR #511, opened 2020-04-08
Encode view state to URL fragment

Hyperkid123:
### Changes
- added hash to URL which represents the filter state of certain pages
  - allows for sharing exact view-state via URL
  - for example, you can create a filter for completed orders with a specific user and bookmark it
- persistent URL is applied on portfolios, portfolio items (of some portfolio), orders and products list screens
- it can be added to any screen by adding some flags to redux action creators
- the view state is encoded via `atob (btoa)` and then added to hash part of URL. It is not sent to servers and it's not limited by 2083 character limit (give or take, depends on ob browser).

persitent-url-filters

### DIFF
--- a/src/utilities/view-state-middleware.js
+++ b/src/utilities/view-state-middleware.js
@@ -1,0 +1,21 @@
+import { encodeState } from '../routing/uri-state-manager';
+
+const viewStateMiddleware = () => (dispatch) => (action) => {
+  if (
+    action.type.match(/_FULFILLED$/) &&
+    action?.payload?.meta &&
+    action?.meta?.storeState
+  ) {
+    window.location.hash = encodeState(
+      {
+        ...action.meta,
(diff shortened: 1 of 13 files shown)


Hyperkid123:
cc @rvsia

src/utilities/view-state-middleware.js
```diff
@@ -0,0 +1,23 @@
+import { encodeState } from '../routing/uri-state-manager';
+
+const viewStateMiddleware = () => (dispatch) => (action) => {
+  if (
+    action.type.match(/_FULFILLED$/) &&
+    action.payload &&
+    action.payload.meta &&
```
rvsia, line 7: you can use `?` chaining here

lgalis:
@Hyperkid123 - will merge after the rebase

PR #515, opened 2020-04-14
Fix button variants in edit product form.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1433

### Before
screenshot-ci cloud redhat com-2020 04 14-09_28_41

### After
screenshot-ci foo redhat com_1337-2020 04 14-09_27_59

cc @sbuenafe-rh

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3525,7 +3525,7 @@
     "@emotion/css": {
       "version": "10.0.27",
       "resolved": "https://registry.npmjs.org/@emotion/css/-/css-10.0.27.tgz",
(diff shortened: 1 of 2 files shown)


sbuenafe-rh:
@Hyperkid123 Looks good, but the Reset button looks like it is disabled. Can you use the secondary button styling for it? Thanks!

sbuenafe-rh:
@Hyperkid123 after taking a little bit more look at this Edit page, we don't need the Reset button. Canceling the edit state can take place of resetting. I would remove the reset button from the form.

Hyperkid123:
@sbuenafe-rh the reset button is gone.

PR #516, opened 2020-04-14
Update ddf to v2.

Hyperkid123:
It can wait for weeks. Just putting it out there so we won't forget: https://github.com/RedHatInsights/catalog-api/pull/671

Currently, there are components with old keys in the component mapper to serve as a bridge between the new version of DDF and older schemas in DB.

Redundant components and validators should be removed after DB migration.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3272,9 +3272,9 @@
       }
     },
     "@babel/runtime-corejs2": {
-      "version": "7.8.7",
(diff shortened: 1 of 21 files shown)


Hyperkid123:
@rvsia. Yes, I need to sometimes pre-fetch paginated initial values options in load options. 
We also sometimes add empty option because we were not allowed to use the clearable option in some places. I prefer to do that in one place.

PR #523, opened 2020-04-15
Fix Order icon in mobile layout

epwinchell:
jira: https://projects.engineering.redhat.com/browse/SSP-1436
follow-up to: https://github.com/RedHatInsights/catalog-ui/pull/512

Changes introduced in the PR above allowed for cropping of product images when necessary, but also severely cropped them in the mobile layout where there's room. This PR excludes the "shrink" class from the mobile layout, resolving the issue.

Old (mobile)
Screen Shot 2020-04-16 at 3 03 31 PM

New (mobile)

Screen Shot 2020-04-16 at 3 00 41 PM

Old (desktop)

Screen Shot 2020-04-16 at 3 03 19 PM

New (desktop)

Screen Shot 2020-04-16 at 3 00 10 PM

### DIFF
--- a/src/presentational-components/styled-components/table.js
+++ b/src/presentational-components/styled-components/table.js
@@ -31,9 +31,11 @@ Tbody.propTypes = {
 export const TableCell = styled(({ shrink, children, ...props }) => (
   <td {...props}>{children}</td>
 ))`
-  vertical-align: middle !important;
-  width: ${({ shrink }) => (shrink ? '10%' : 'initial')};
-  img {
-    object-fit: cover;
+  @media screen and (min-width: 768px) {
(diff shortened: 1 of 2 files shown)


src/presentational-components/styled-components/table.js
```diff
@@ -28,11 +28,10 @@ Tbody.propTypes = {
   children: childrenPropType.isRequired
 };
 
-export const TableCell = styled(({ shrink, children, ...props }) => (
```
lgalis, line 31: This file can stay as is, not sending the shrink parameter in the order-item would be enough.

lgalis:
@epwinchell  - shrink is needed for the column to maintain the current layout

PR #539, opened 2020-04-23
Long Portfolio name breaks the layout

epwinchell:
jira: https://projects.engineering.redhat.com/browse/SSP-1455

Extremely long names (without spaces) cause screen elements on the right to be pushed out of the viewport. Here, I used "overflow-wrap: anywhere", which will first attempt to wrap at a word break, failing that, it will break extremely long words where needed.  Since the breadcrumb and title are able to accommodate wrapping, it seemed like a better  approach than using an ellipsis.

/cc @sbuenafe-rh

Old
Screen Shot 2020-04-23 at 3 44 09 PM

New
Screen Shot 2020-04-23 at 3 43 25 PM
Screen Shot 2020-04-23 at 4 14 37 PM

### DIFF
--- a/src/global-styles.js
+++ b/src/global-styles.js
@@ -59,12 +59,16 @@ const GlobalStyle = createGlobalStyle`
   flex-grow: 1;
 }
 
-a.pf-c-breadcrumb__item {
-  cursor: pointer;
-  >* {
+.pf-c-breadcrumb__list {
+  overflow-wrap: anywhere;
+  a.pf-c-breadcrumb__item {
     cursor: pointer;
+    >* {
+      cursor: pointer;
+    }
   }
 }
+
 .pf4-hidefield-overlay{
   .hide-indicator {
     z-index: 1;
--- a/src/presentational-components/styled-components/toolbars.js
+++ b/src/presentational-components/styled-components/toolbars.js
@@ -21,6 +21,7 @@ export const TopToolbarWrapper = styled.div`
   }
   h2 {
     margin-bottom: 0 !important;
+    overflow-wrap: anywhere;
   }
   .top-toolbar-title {
     min-width: 200px;


sbuenafe-rh:
@epwinchell yes, thank you for bringing this up. Portfolio names should have a cap on the number of characters. What that limit is? I would have to do more investigating for a best practice. 

The title area on a portfolio card that truncates with the ellipses should have a tooltip that displays the entire portfolio name.

gmcculloug:
@sbuenafe-rh fyi, I found the other day that Tower Job Templates have a max name limit of 512 characters.  I realize that relates more to Products than Portfolios, but thought it might be useful information here.

epwinchell:
> The title area on a portfolio card that truncates with the ellipses should have a tooltip that displays the entire portfolio name.

@sbuenafe-rh
Screen Shot 2020-04-24 at 10 17 31 AM

sbuenafe-rh:
@epwinchell and @gmcculloug I just found out for Ansible namespaces it's 2-64 character limit and collection names are 64 max. The limit content within a collection is unknown.

sbuenafe-rh:
> > The title area on a portfolio card that truncates with the ellipses should have a tooltip that displays the entire portfolio name.
> 
> @sbuenafe-rh
> Screen Shot 2020-04-24 at 10 17 31 AM

@epwinchell is that the PF4 tooltip? https://www.patternfly.org/v4/documentation/react/components/tooltip

epwinchell:
> @epwinchell is that the PF4 tooltip? 
@sbuenafe-rh It doesn't look quite right. I'll investigate.

sbuenafe-rh:
@epwinchell lets go with a max of 64 characters for the portfolio name. 

@gmcculloug do we have a minimum char for portfolio name?

lgalis:
Created a new Jira card for the tooltip and product/portfolio name limit:
https://projects.engineering.redhat.com/browse/SSP-1465

gmcculloug:
@sbuenafe-rh The API enforces that the name cannot be blank, there is no minimum length check beyond that.

If we are enforcing length (min = 1, max = 64) on name we need to add that to the API and I believe exposed those min/max values through `openapi.json`.  @mkanoor is that correct?

cc @eclarizio

gmcculloug:
@mkanoor @eclarizio Please review previous comment.

mkanoor:
@gmcculloug @eclarizio 
Currently there are no limits on portfolio.name and portfolio_item.name. If we want to add limits we can do that or in the UI have a max limit for a text field and display ellipsis after the max display limit.

gmcculloug:
@mkanoor If we are going to enforce a limit on the values it has to be enforced from the API, not just the UI.

For UI support would this also be exposed through the `openapi.json` spec?

I'll open a Jira issue for the new sprint (starting today) that limits the Portfolio name to 64 characters per https://github.com/RedHatInsights/catalog-ui/pull/539#issuecomment-619094834 above.

PR #550, opened 2020-04-30
Details screens responsiveness follow-up

epwinchell:
jira: https://projects.engineering.redhat.com/browse/SSP-1457
follow-up to: https://github.com/RedHatInsights/catalog-ui/pull/531

The last PR (https://github.com/RedHatInsights/catalog-ui/pull/531)  made adjustments to the left column on the Service Offering & Portfolio screens to make them responsive and removed the ellipses, allowing wrapping at word breaks.

This PR forces long names without spaces to break where needed and adds proper padding between columns.

Before:
Screen Shot 2020-04-30 at 11 05 47 AM

After:
Screen Shot 2020-04-30 at 11 22 39 AM

### DIFF
--- a/src/global-styles.js
+++ b/src/global-styles.js
@@ -12,6 +12,10 @@ const GlobalStyle = createGlobalStyle`
   margin-bottom: 2px !important;
 }
 
+.overflow-wrap {
+  overflow-wrap: anywhere;
+}
+
 .orders-list {
   background-color: var(--pf-global--BackgroundColor--100)
 }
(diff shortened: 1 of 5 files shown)


src/global-styles.js
```diff
@@ -12,6 +12,10 @@ const GlobalStyle = createGlobalStyle`
   margin-bottom: 2px !important;
 }
 
+.overflow-wrap {
+  overflow-wrap: anywhere;
```
Hyperkid123, line 16: We might potentially have an issue with browser support. I believe Safari (the new IE) does not support the `anywhere` wrap.

We might want to consider adding some fallback with `@supports` for those who still use safari.

lgalis: @Hyperkid123, @epwinchell - I did not see this comment before merging - this can be addressed in a separate PR

epwinchell: @Hyperkid123 Will do. Thanks.

PR #553, opened 2020-05-01
Set max length for the portfolio name as specified in the openapi schema

lgalis:
Fixes https://projects.engineering.redhat.com/browse/SSP-1482

### DIFF
--- a/src/forms/portfolio-form.schema.js
+++ b/src/forms/portfolio-form.schema.js
@@ -32,8 +32,6 @@ export const createPortfolioSchema = (
   openApiSchema,
   portfolioId
 ) => {
-  console.log('DEBUG - openApiSchema: ', openApiSchema);
-
   return {
     fields: [
       {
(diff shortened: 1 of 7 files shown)


src/forms/portfolio-form.schema.js
```diff
@@ -3,6 +3,8 @@ import { componentTypes } from '@data-driven-forms/react-form-renderer';
 import asyncFormValidator from '../utilities/async-form-validator';
 import { fetchPortfolioByName } from '../helpers/portfolio/portfolio-helper';
 
+const DEFAULT_MAX_LENGTH = 65;
```
eclarizio, line 6: Is this just a js thing or is this a typo? [Our current max length is 64](https://github.com/RedHatInsights/catalog-api/blob/master/public/doc/openapi-3-v1.0.json#L2845

lgalis: typo

eclarizio: Ah ok, well, I guess that then begs the question is there some way to actually get this value from the openapi schema? If we update this value in the future, you're going to have to come in here and re-update this value again, which doesn't seem like a great thing to have to do.

lgalis: This PR gets the value from the openApi schema - the default is there only if there is no value in the openaAPI schema

eclarizio: Got it, sounds good.

PR #555, opened 2020-05-04
Fixed missing filtering options after adding/editing portfolio.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1486

### Changes
- added the view state to add/update portfolio API call to preserver filtering and pagination settings

### DIFF
--- a/src/presentational-components/portfolio/porfolio-card.js
+++ b/src/presentational-components/portfolio/porfolio-card.js
@@ -46,6 +46,7 @@ const HeaderActions = ({
           <CatalogLink
             searchParams={{ portfolio: portfolioId }}
             pathname={SHARE_PORTFOLIO_ROUTE}
+            preserveHash
           >
             Share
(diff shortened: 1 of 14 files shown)


src/helpers/shared/user-login.js
```diff
@@ -26,10 +26,7 @@ const axiosInstance = axios.create({
 
 const resolveInterceptor = (response) => response.data || response;
 const errorInterceptor = (error = {}) => {
-  const requestId = error.response?.headers['x-rh-insights-request-id'];
```
lgalis, line 29: this was added to the notifications frontend-component, but it had the incorrect parameter for the request id. 
We should leave this in place until my PR for the frontend components is merged and released.

Hyperkid123: Right, I've removed it because of tests debugging I forgot to restore the line. :+1:

PR #581, opened 2020-05-15
Fixed portfolio reload after copy product action.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1533

### Fixes
- The portfolio was correctly updated but not stored in the redux store. That caused the old data to be shown even though the URL had the correct redirect information
- Also fixed a typo in disabling submit prop on the form

copy-redirect


Hyperkid123:
I Just ran npm install. I Don't know why exactly there are changes. Maybe security vulnerability fix or we forgot to update the lock file in the past? I will try to run the install again in the morning.

Hyperkid123:
Ok, I've cleaned my npm cache rebased on master lock file and the diff is gone :shrug:. It seems to be some random fluke.

PR #590, opened 2020-05-26
Add a hash to JS bundles

Hyperkid123:
In preparation for service workes and browser caching and pre-loading of the assets, we need to ensure that w have unique asset names in order to be able to use the most aggressive and efficient cache strategy.

The webpack `contentHash` generates a unique hash based on the file contents. This should allow us to use the `CacheFirst` service worker cache strategy.

I've also had to adjust the spadnx confix because we were missing the root override `/anasible/catalog` and it was serving the old `App.js` and `index.html` directly from ci.cloud and not the localhost.

### We no longer need to develop on /beta version of the. In order to do that just add `BETA=true` before the UI start script. It was done automatically before.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -37,8 +37,8 @@ const commonConfig = {
     }
   },
   output: {
-    filename: 'js/[name].js',
-    chunkFilename: 'js/[name].js',
+    filename: 'js/[name].[contenthash].js',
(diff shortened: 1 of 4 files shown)


Hyperkid123:
I have found some issues with serving the hashed files when developing locally. The proxy is not redirecting the asset requests to the localhost. We will probably need some additional changes to the spandx config. Looking at that now.

PR #596, opened 2020-05-26
eliminate unneeded scrollbar from modals

epwinchell:
jira: https://projects.engineering.redhat.com/browse/SSP-1529

This PR removes the vertical scrollbar present in all modal bodies.

Before: 
Screen Shot 2020-05-12 at 7 07 11 AM

After:
Screen Shot 2020-05-26 at 2 23 54 PM

### DIFF
--- a/src/global-styles.js
+++ b/src/global-styles.js
@@ -94,6 +94,11 @@ a.pf-c-breadcrumb__item {
     z-index: 1;
   }
 }
+
+.pf-c-modal-box__body {
+  overflow-y: hidden !important;
+}
+
 /**
 * End of PF4 fixes
 */


src/global-styles.js
```diff
@@ -94,6 +94,11 @@ a.pf-c-breadcrumb__item {
     z-index: 1;
   }
 }
+
+.pf-c-modal-box__body {
+  overflow-y: hidden !important;
```
Hyperkid123, line 99: This will cause an issue on small devices or when the form has many fields (a big list of configuration when ordering job template).

Users won't be able to scroll down to the overflowing content.

screenshot-ci cloud redhat com-2020 05 27-08_41_54

The solution here is directly tied to the PF modal HTML hierarchy. And I am not sure if we can simply fix it by changing CSS until they give us full control over how we can build these dialogs.

PR #598, opened 2020-05-27
Add workbox service worker

Hyperkid123:
wait for #590 

### Changes
- Added simple service worker to cache static assets
- Service worker should significantly improve the performance and also reduce load times when accessing the page and its assets offline or in low bandwidth environment (API will still not work, we need to implement caching our selves and that will require the server push feature)
- **The service worker registration is a nonessential operation and will not cause any app issues if the registration fails**
- the current implementation only pre-caches JS files and does nothing else
- This was created to mainly test and discover limitations of the platform capabilities when it comes to pre-caching static assets
- IF you want to test it locally, you will need to run the UI outside of the proxy in non-https env
  - service worker security policy does not allow self-signed certificates and the proxy will cause issues with different script origin which is forbidden by the service worker policy in order to prevent any XSS attacks from severe workers
  - you will have to mock several API endpoints and, running the API locally is the best bet in order to do some testing


Hyperkid123:
All right. Again we don't have to worry if the registration fails.

PR #629, opened 2020-06-18
Add platform detail toolbar loader.

Hyperkid123:
### Changes
- added small toolbar placeholder when switching between platform templates and inventories
  - removes the ugly popping of the toolbar after the JS chunk is loaded for the first time
  - transition is a bit easier on the eye

screenshot-ci foo redhat com_1337-2020 06 18-15_49_00

### DIFF
--- a/src/presentational-components/shared/loader-placeholders.js
+++ b/src/presentational-components/shared/loader-placeholders.js
@@ -23,6 +23,7 @@ import {
   Button
 } from '@patternfly/react-core';
 import styled, { keyframes } from 'styled-components';
+import { StyledToolbar } from '../styled-components/toolbars';
 
 const wave = keyframes`
   0% {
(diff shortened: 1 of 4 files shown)


src/presentational-components/shared/loader-placeholders.js
```diff
@@ -215,3 +216,9 @@ ListLoader.defaultProps = {
 };
 
 export const OrderDetailToolbarPlaceholder = () => <Skeleton height={70} />;
+
+export const PlatformToolbarPlaceholder = () => (
+  <StyledToolbar className="pf-u-pr-lg pf-u-pl-lg pf-u-pb-md pf-u-pt-md">
```
rvsia, line 221: "pf-u-pr-lg pf-u-pl-lg pf-u-pb-md pf-u-pt-md"

 the first two could be collapsed into one

"pf-u-p-lg pf-u-pb-md pf-u-pt-md"

but not sure if the PF css allows this.

src/presentational-components/shared/loader-placeholders.js
```diff
@@ -215,3 +216,9 @@ ListLoader.defaultProps = {
...
+    <Skeleton height={36} width="300px" />
```
rvsia, line 222: just nitpick: one is number and the second is string. :D

src/presentational-components/shared/loader-placeholders.js
Hyperkid123: Hey, I am happy that it even works :smile: who knows what it will do when we try to be clever.

PR #647, opened 2020-07-08
Fixed outdated toolbar layout.

Hyperkid123:
There was an issue with underlying PF toolbar components. [Fix for platform toolbars](https://github.com/RedHatInsights/catalog-ui/pull/645) distorted the rest of the toolbars. Change to the actual DOM is necessary to make it work on all pages.

### Changes
- use new Toolbar component composition
- added wrap property to toolbar component
- added wrap property to toolbar group component (no idea why toolbar group is not wrapped by default)

### Before
screenshot-ci cloud redhat com-2020 07 08-11_10_06

### After
screenshot-ci foo redhat com_1337-2020 07 08-11_10_28

### DIFF
--- a/src/presentational-components/styled-components/toolbars.js
+++ b/src/presentational-components/styled-components/toolbars.js
@@ -1,16 +1,23 @@
+import React from 'react';
 import styled from 'styled-components';
-import { Level } from '@patternfly/react-core';
+import { Level, Toolbar, ToolbarGroup } from '@patternfly/react-core';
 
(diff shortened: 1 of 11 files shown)


lgalis:
@Hyperkid123 - looks good, I will merge after the conflicts are resolved.

Not for this PR - the Front end comp. toolbar buttons collapses  under a kebab on a mobile screen - should probably have the same functionality in the Catalog toolbars.

PR #654, opened 2020-07-10
Translate catalog

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1658

I had to skip one integration test. Issues are cause the `react-final-form` dependency. It tries to do updates on unmounted components with async validation and causes issues in non-browser environments.

(We will probably use custom state management library at some point)

`formatMessage` cannot be used for notifications because the text is not then parsed in a portal in a test environment. This causes issues in interactive notifications. FormattedMessage works like before.

### TODO
- [x] translate static strings
- [x] translate dynamic strings where it makes sense (like order status)
- [x] translate form statically defined schemas
  - will require some clever system to get the translation context outside react component
- [x] deduplicate common strings
- [x] think of some nice ids for messages
- [x] translate notification texts defined in redux actions
- [x] add tests for schema translator
- [x] fix broken texts
  - most components will miss the i18l context and, should be easy enough to add it
- [x] move towards imperative API instead of components (better performance)
- [x] make the code pretty


Hyperkid123:
These notifications will need a bit more work. I am thinking about slapping some middleware which would add an additional argument to the actions and use one global object outside of the redux. But I would do that as a follow-up. This PR is way too big already.

PR #663, opened 2020-07-17
Switch order detail to tabs layout

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1671

### Changes
- use tab layout instead of the vertical layout
- content is now in one big card this is just **temporary** because there is a follow-up task
- I was unable to use the official PF component due to some issues
  - does not support disabled tabs
  - when adding a custom class to the tab item to disable pointer events, the styling breaks (related issue: https://github.com/patternfly/patternfly-react/issues/4567)

screenshot-ci foo redhat com_1337-2020 07 17-11_06_39

cc @fionayhlin


gmcculloug:
cc @sbuenafe-rh

lgalis:
@Hyperkid123 , @sbuenafe-rh
If we cannot use the PF tab because it does not support disabled tabs - have we considered having all tabs enabled with empty content?

Hyperkid123:
@lgalis I was chatting with @sbuenafe-rh and the nav will receive probably some additional changes.

sbuenafe-rh:
@Hyperkid123 @lgalis if disabled tabs aren't allowed then, yes we'll have to supply empty states to those tabs. @fionayhlin is looking into if PF will have a disabled state for tabs in the future.

PR #665, opened 2020-07-20
Order detail cards

Hyperkid123:
wait for #663 
jira: https://projects.engineering.redhat.com/browse/SSP-1673

###Changes
- update the approval tab according to new design
screenshot-ci foo redhat com_1337-2020 07 20-14_11_50

screenshot-ci foo redhat com_1337-2020 07 20-14_09_23

screenshot-ci foo redhat com_1337-2020 07 20-14_12_01

cc @sbuenafe-rh @fionayhlin

I do have a couple of questions for our UX team.
- I see that the updated date for approval on mocks is using the exact value, but we have been using the relative date everywhere else. So should it bee relative or exact
- There was an icon for the completed request. Do we have icons for the rest of the request states?
- And what about the lifecycle tab? Currently, there is just the link. Are there any changes to that tab?

### DIFF
--- a/src/helpers/order/order-helper.js
+++ b/src/helpers/order/order-helper.js
@@ -3,15 +3,21 @@ import {
   getAxiosInstance,
   getPortfolioItemApi,
   getOrderApi,
-  getOrderItemApi
+  getOrderItemApi,
+  getGraphqlInstance
 } from '../shared/user-login';
-import { CATALOG_API_BASE, SOURCES_API_BASE } from '../../utilities/constants';
+import {
+  CATALOG_API_BASE,
+  SOURCES_API_BASE,
+  APPROVAL_API_BASE
+} from '../../utilities/constants';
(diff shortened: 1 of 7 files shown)


sbuenafe-rh:
@Hyperkid123 

> I see that the updated date for approval on mocks is using the exact value, but we have been using the relative date everywhere else. So should it bee relative or exact
There was an icon for the completed request. Do we have icons for the rest of the request states?
And what about the lifecycle tab? Currently, there is just the link. Are there any changes to that tab?

Yes, so we do use relative time everywhere, but in the activity area since it is a log, we should be using the absolute time stamp. 

For the reminder of the request states icon styles -- TBD on the icons UXD is looking into all statuses.

Lifecycle link tab, we're still discussing what's going on here.

Hyperkid123:
@sbuenafe-rh its finished then. Just waiting for the PR dependency.

PR #678, opened 2020-07-23
Share modal refactor

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1681

### Changes
- the group list is now dynamic
  - can add multiple groups at once
  - user must use the `+` button to add the group
- new icon for group deletion
- submit button is disabled until the form is in `dirty` state
 

share

### DIFF
--- a/src/forms/form-fields/share-group-select.js
+++ b/src/forms/form-fields/share-group-select.js
@@ -1,77 +1,127 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { Grid, GridItem, Button, Level, Tooltip } from '@patternfly/react-core';
+import PlusIcon from '@patternfly/react-icons/dist/js/icons/plus-icon';
 import { InternalSelect } from '@data-driven-forms/pf4-component-mapper/dist/cjs/select';
-import { Grid, GridItem } from '@patternfly/react-core';
+import FieldArray from '@data-driven-forms/react-form-renderer/dist/cjs/field-array';
+
 import asyncFormValidator from '../../utilities/async-form-validator';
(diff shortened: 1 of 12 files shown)


src/forms/form-fields/share-group-select.js
```diff
@@ -1,77 +1,127 @@
...
+              isDisabled={!group || !permission}
+              onClick={() => {
+                addGroup({
+                  groupName: group.label,
+                  group_uuid: group.value,
+                  permissions: permission
+                });
+                setPermission(undefined);
+                setGroup(undefined);
+                setResetGroup((prev) => prev + 1);
```
rvsia, line 100: use `useReducer` instead? :thinking:

Hyperkid123:
I will squash the messages as well.

Hyperkid123:
I've updated the sharing success message. We could add the numbers for share/unshare but honestly, I feel like this message is far better, considering that when user only updates the actually shares/unshares the portfolio which is very confusing.

screenshot-ci foo redhat com_1337-2020 07 27-10_00_54

gmcculloug:
I like the single message but want to confirm with @sbuenafe-rh.  Also, what does the failure message look like?

Hyperkid123:
The failure message is a problem. Since we have to create one API call per each share/unshare the user will get an error message only for the API call that failed. So you will get whatever error message the API returns.

In addition, if one of the calls fails, the rest of share/unshare actions will be persisted. Obviously we can't rollback the database form UI at that point.

lgalis:
@Hyperkid123 - I will merge this PR, as it works well - and the issue in my comment above will be fixed in #682

PR #682, opened 2020-07-27
Fix copy select value

Hyperkid123:
jira:
- https://projects.engineering.redhat.com/browse/SSP-1714
- https://projects.engineering.redhat.com/browse/SSP-1716

### Fixes
- spacebar key event is normal used to close/open the select menu (aria accessibility standard)
  - the input was propagating the event to the toggle root element which causes unwanted menu toggle
  - the keyDown event is no longer propagated to its parents (DDF PR: https://github.com/data-driven-forms/react-forms/pull/660)
- The text component showing the next name causing form re-initialization which deletes the value from the select component
  - the next name component is now directly connected to the form state and has its internal logic to get the next name instead

select-pf

### DIFF
--- a/src/test/integration/portfolio-item-integration.test.js
+++ b/src/test/integration/portfolio-item-integration.test.js
@@ -356,7 +356,7 @@ describe('Integration tests for portfolio items', () => {
       .onGet(
         `${CATALOG_API_BASE}/portfolio_items/source-offering-1/next_name?destination_portfolio_id=123`
       )
-      .replyOnce(200, { next_name: `Copy of ${addedPortfolioItem.name}` });
+      .reply(200, { next_name: `Copy of ${addedPortfolioItem.name}` });
     mockApi
       .onGet(
         `${CATALOG_API_BASE}/portfolios?filter[name][contains_i]=&limit=100&offset=0`
--- a/package-lock.json
+++ b/package-lock.json
@@ -1237,18 +1237,18 @@
       }
     },
     "@data-driven-forms/pf4-component-mapper": {
-      "version": "2.8.4",
-      "resolved": "https://registry.npmjs.org/@data-driven-forms/pf4-component-mapper/-/pf4-component-mapper-2.8.4.tgz",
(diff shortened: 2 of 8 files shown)


Hyperkid123:
Yeah :disappointed: that is the default behavior. Let's see what I can do about it.

Hyperkid123:
@lgalis the filter is no longer applied when the menu is opened

Hyperkid123:
@lgalis did you re-install dependencies? I don't see the error. Or does it occur in a specific scenario?

src/test/integration/portfolio-item-integration.test.js
```diff
@@ -356,7 +356,7 @@ describe('Integration tests for portfolio items', () => {
       .onGet(
         `${CATALOG_API_BASE}/portfolio_items/source-offering-1/next_name?destination_portfolio_id=123`
       )
-      .replyOnce(200, { next_name: `Copy of ${addedPortfolioItem.name}` });
+      .reply(200, { next_name: `Copy of ${addedPortfolioItem.name}` });
```
rvsia, line 359: Is `mock.reset();` called after this? I can imagine that changing from `once` could affect other test files.

Hyperkid123: It won't since this endpoint must always return the same value. Now its called multiple times since it's handled by the "spy" component.

Hyperkid123:
@rvsia 
> When you select an option, the select is loading 

This is actually how the DDF is implemented. We can review it over there.

> Sometimes styling goes totally crazy

Seems like we need max-height for very large lists. 

Edit: The list gets massive while the select is still loading. Can you confirm?

rvsia:
@Hyperkid123 Yeah, the loading seems to cause the issue.

Hyperkid123:
Ok I will take a look. Hopefully, this nightmare will end soon.

Hyperkid123:
@rvsia ok the menu warping was solved as well.

PR #683, opened 2020-07-27
Give product set approval component higher priority.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1715

### Fixes
- the route for edit workflows was "eaten" by different component so the form did not even mount
  - component with the modal has now the top priority
- removed the buttons route restrictions
  - toolbar buttons are no longer hiding if the user is another route than `/portfolio/portfolio-item`


Hyperkid123:
Yep there should be another ticket for this issue in the backlog. Should I include a fix into this PR?

lgalis:
@Hyperkid123 - I will merge the PR as is - the other issue can be fixed in a separate PR

PR #684, opened 2020-07-28
Set selectedPortfolio when changing portfolio query param

rvsia:
Fixes https://projects.engineering.redhat.com/browse/SSP-1676

- `share` and `setApproval` modals depend on `selectedPortfolio`, however this value was not filled in the list. So now the `routes` component check if there is `portfolio` query and if no portfolio is selected, it will select it automatically + if the portfolio is not in the store already, it's fetched (i.e. on access via URL).
- `removeQuery` prop does not exist, changed to `removeSearch`

**Before**

image

**After**

image

### DIFF
--- a/src/smart-components/dialog-routes/portfolio-routes.js
+++ b/src/smart-components/dialog-routes/portfolio-routes.js
@@ -1,4 +1,4 @@
-import React, { lazy } from 'react';
+import React, { lazy, useEffect } from 'react';
 import { Switch, Route } from 'react-router-dom';
 import {
   ADD_PORTFOLIO_ROUTE,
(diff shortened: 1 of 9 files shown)


Hyperkid123:
(did not notice the WIP and failing travis)

rvsia:
@Hyperkid123 ready to review!

src/smart-components/dialog-routes/portfolio-routes.js
```diff
@@ -55,21 +59,33 @@ const AddPortfolioModal = lazy(() =>
...
-        state?.portfolioReducer?.portfolioItem?.portfolioItem?.metadata
-          ?.user_capabilities
-    })
-  );
+  const {
+    portfolioItemId,
+    portfolioItemUserCapabilities,
+    portfolios,
+    selectedPortfolio
+  } = useSelector((state) => ({
```
Hyperkid123, line 67: this selector is now probably updated all the time and re-rendering the whole thing. Could you split it into one selector per value and add shallow equal?

PR #691, opened 2020-07-29
Update add portfolio notification

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1686

### Changes
- notification is translated
- the portfolio name is **bold**

screenshot-ci foo redhat com_1337-2020 07 29-13_25_10

### DIFF
--- a/src/smart-components/portfolio/add-portfolio-modal.js
+++ b/src/smart-components/portfolio/add-portfolio-modal.js
@@ -34,7 +34,16 @@ const AddPortfolioModal = ({ removeQuery, closeTarget, viewState }) => {
 
   const onAddPortfolio = async (data) => {
     setSubmitting(true);
-    const newPortfolio = await dispatch(addPortfolio(data));
+    const notification = {
+      variant: 'success',
+      title: formatMessage(portfolioMessages.addSuccessTitle),
+      description: formatMessage(portfolioMessages.addSuccessDescription, {
+        name: data.name,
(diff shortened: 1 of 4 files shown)


src/smart-components/portfolio/add-portfolio-modal.js
```diff
@@ -34,7 +34,16 @@ const AddPortfolioModal = ({ removeQuery, closeTarget, viewState }) => {
...
   const onAddPortfolio = async (data) => {
     setSubmitting(true);
-    const newPortfolio = await dispatch(addPortfolio(data));
+    const notification = {
+      variant: 'success',
+      title: formatMessage(portfolioMessages.addSuccessTitle),
+      description: formatMessage(portfolioMessages.addSuccessDescription, {
+        name: data.name,
+        // eslint-disable-next-line react/display-name
+        b: (chunks) => <b>{chunks}</b>
```
rvsia, line 43: I think you should use `key` here, otherwise intl will be mad. :crab:

Hyperkid123: Oh yeah, I hate that. Adding arbitrary keys to something we should have to :angry:

Hyperkid123: I think i will make a wrapper around that function and fix it in all places afterwards (I've added the key here but there is a lot of places with the same issue)

PR #693, opened 2020-07-29
Update edit workflow process modal

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1682

### Changes
- existing approval workflows are now inside a chip group
- existing approval workflows are no longer an initial value of the select
- select options are disabled if the option is present in the chip group
- the submit callback only sends necessary workflow ids
  - if the user unlinks workflow and then picks the same from the select, that id is not passed to the API call.

processes


Hyperkid123:
@rvsia I would guess not. We are using on mouse event which is triggered even though all pointer-events on the element are disabled. we should check that out in DDF.

PR #690, opened 2020-07-29
Add advanced filtering to portfolios screen

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1725, https://projects.engineering.redhat.com/browse/SSP-1692

### Changes
- use primary toolbar for portfolios
- add multi-filtering support
  - user can filter by name or owner
- add sorting support
  - only by one attribute since that is what current API supports
  - sortable parameters are name, owner, created_at, updated_at

- [x] translate strings

advanced-filter-portfolios

PR #700, opened 2020-07-30
Copy created messages to states messages

Hyperkid123:
Introduced in the advanced filter for portfolios or #690 

jira: https://projects.engineering.redhat.com/browse/SSP-1726

We have to be very careful when moving messages around. This was a dynamically assigned value to the format message API. I think I will create a wrapper around that function to add keys to rich text translation and a fallback so it does not cause runtime crashes if the definition was not found.

### DIFF
--- a/src/messages/states.messages.js
+++ b/src/messages/states.messages.js
@@ -1,3 +1,5 @@
+import labelMessages from './labels.messages';
+
 const { defineMessages } = require('react-intl');
 
 const statesMessages = defineMessages({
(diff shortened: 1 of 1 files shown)


src/messages/states.messages.js
```diff
@@ -1,3 +1,5 @@
+import labelMessages from './labels.messages';
```
rvsia, line 1: mixing `require`/ES `import`

Hyperkid123:
@rvsia it fails because we send an `undefined` and the format messages do not have the message ID. The messages are picked based on API response status (an object with status value and that is remapped to the defined message). It dies crash throwing an error. More details in another PR

src/messages/states.messages.js
Hyperkid123: ah vs code was confused i guess

PR #701, opened 2020-07-30
Custom format messages

Hyperkid123:
### Changes
- replace `formatMessage` with custom `useFormatMessage` hook
  - uses react int formater obviously, but it changes the params to fix some of its issues.
  - adds a fallback message to prevent runtime crashes for undefined or incorrectly defined messages
  - fixes missing key react error in rich text messages by adding a Fragment wrapper

(most of the changes in this PR are just import changes)

### DIFF
--- a/src/utilities/use-format-message.js
+++ b/src/utilities/use-format-message.js
@@ -1,12 +1,24 @@
+import React, { Fragment } from 'react';
 const { useIntl } = require('react-intl');
 
 const useFormatMessage = () => {
   const { formatMessage } = useIntl();
-  return (message, values) => {
+  return (message, values = {}) => {
+    const internalValues = Object.entries(values).reduce(
+      (acc, [key, value]) => {
+        return {
+          ...acc,
+          [key]:
+            typeof value === 'function'
+              ? (chunks) => <Fragment key={key}>{value(chunks)}</Fragment>
+              : value
+        };
+      },
+      {}
+    );
     try {
-      return formatMessage(message, values);
+      return formatMessage(message, internalValues);
     } catch (error) {
-      console.error(error);
       return `Unable to translate message. Definition: ${message}, values: ${values}, intl-error: ${error}`;
     }
   };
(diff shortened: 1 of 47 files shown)


src/utilities/use-format-message.js
```diff
@@ -0,0 +1,27 @@
...
+
+const useFormatMessage = () => {
+  const { formatMessage } = useIntl();
+  return (message, values = {}) => {
+    const internalValues = Object.entries(values).reduce(
+      (acc, [key, value]) => {
+        return {
+          ...acc,
+          [key]:
+            typeof value === 'function'
```
rvsia, line 12: What if you want to put an element there?

Hyperkid123: Then its returned as it was. You can't inject key to cloned element since its not a prop anyway. I think we can make this a convention in our apps to just use function components.

rvsia: You can still wrap an element in the fragment too.

Hyperkid123: Yes, you can but detecting an element is fairly ambiguous. The react element does not work very well if you remember when we have tried that in DDF. It all comes down to if you can call the value like `value()` or like `<value />`. We might just try the latter and see if it works though. :thinking: gimme a moment.

rvsia: I think checking `typeof x === 'object'` should be enough for this case. (As you can't put a standard object there.)

Hyperkid123: True and even if it does fail, there is still the try-catch block which will catch the error anyway:shrug:. Here I am trying to think of stuff that won't happen...

Hyperkid123: Nah it won't work. We are starting to get to some stupidly large overhead for what we try to achieve. The chunks parm is not your standard component props object.

```jsx
            typeof value === 'function' || typeof value === 'object'
              ? (chunks) => {
                  const Component = value;
                  return (
                    <Fragment key={key}>
                      <Component>
                        {chunks.map((chunk, index) => (
                          <Fragment key={index}>{chunk}</Fragment>
                        ))}
                      </Component>
                    </Fragment>
                  );
                }
              : value
```

this is just too much to make such a simple thing working. If we won't use class components (which we are not) we can do wich just the function type.

PR #708, opened 2020-08-02
[Prod-Stable] Create manifest file from prod-stable branch

gmcculloug:


Hyperkid123:
Just a note. We could modify the manifest script to exclude devDependencies since they do not make it to the production build. And I would guess that we are concerned about the production version of the build?

I don't know what are the exact environments though.

gmcculloug:
@Hyperkid123 yes, these would be focused on what gets deployed to production.  The command I'm using is here if you want to take a look and advise: https://gist.github.com/gmcculloug/b4676879d7db32de52223270b4e8491c

Hyperkid123:
Nice. I will play around with it so we can reduce the list a bit next time we update it.

PR #709, opened 2020-08-04
Tag objects with processes

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1600

Turns out that the approval process modal and order process modals and form schemas are exactly the same apart from some config variables and other parameters. I will merge them into one configurable component.

It has the same functionality as the set approval process modal.

screenshot-ci foo redhat com_1337-2020 08 04-12_25_32

### TO DO
- [x] order process tagging for portfolios
- [x] order process tagging for portfolio items
- [x] consolidate approval and order processes components into one
- [x] tests


src/presentational-components/shared/loader-placeholders.js
```diff
@@ -193,7 +193,7 @@ export const WorkflowLoader = () => {
   const formatMessage = useFormatMessage();
   return (
     <Form>
-      <FormGroup fieldId="2" label="Select workflow">
+      <FormGroup fieldId="2" label="">
```
rvsia, line 196: Is the empty label okay?

src/smart-components/common/edit-approval-workflow.js
```diff
@@ -56,104 +32,46 @@ const EditApprovalWorkflow = ({
...
+        listWorkflowsForObject({
+          objectType,
+          appName: APP_NAME[objectType],
+          objectId: query[querySelector]
+        }).then(({ data }) => data)
+      }
+      loadTags={loadWorkflowOptions}
+      subTitle={formatMessage(approvalMessages.setWorkflow, {
+        // eslint-disable-next-line react/display-name
+        strong: (chunks) => <strong>{chunks}</strong>,
```
rvsia, line 70: In the share modal messages PR I exported this into a separate function. I think it could be moved to a shared file, so there would be no reason to write `// eslint-disable-next-line react/display-name` everywhere.

src/presentational-components/shared/loader-placeholders.js
Hyperkid123: Yes, it's now a common loader. We could customize it but honestly, I don't think that anybody would notice. I would even rather use the spinner.

src/smart-components/common/edit-approval-workflow.js
Hyperkid123: It was the `b` tag right?

Hyperkid123:
@rvsia you are right with the missing notifications. I will add it.

src/smart-components/common/edit-approval-workflow.js
rvsia: Yeah, I think I called it 'bold'.. :smile: You can change it (and the element) to strong, it makes sense in our context.

src/presentational-components/shared/loader-placeholders.js
rvsia: I was speaking more about having an empty prop... according to docs it's not a required attribute.

Hyperkid123:
@rvsia I've added the notification and switched to the 'bold' component.

src/presentational-components/shared/loader-placeholders.js
Hyperkid123: What a feature. Let me remove it.

PR #715, opened 2020-08-06
Add tests for add order process modal

rvsia:
+ add tests for add order process modal
+ small refactoring

@Hyperkid123

### DIFF
--- a/src/messages/forms.messages.js
+++ b/src/messages/forms.messages.js
@@ -22,32 +22,16 @@ const formsMessages = defineMessages({
     defaultMessage: 'Select permission'
   },
   orderProcessName: {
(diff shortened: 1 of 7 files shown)


src/messages/forms.messages.js
```diff
@@ -22,32 +22,16 @@ const formsMessages = defineMessages({
     defaultMessage: 'Select permission'
   },
   orderProcessName: {
-    id: 'formMessages.orderProcessName',
+    id: 'forms.orders.processName',
```
Hyperkid123, line 25: The order and order process are two different things. Order-processes should be the namespace (or something similar).

rvsia: moved!

PR #717, opened 2020-08-10
Add persistent filtering to order processes

Hyperkid123:
Store filtering and pagination settings to order processes URL hash

op-hash


rvsia:
@Hyperkid123 tests failed

PR #718, opened 2020-08-11
Update orders table

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1695

### Changes
- switch the orders table layout once again
- added sorting to the table
  - currently, we can only sort by id, owner, state and created
- the status label is no longer a link

new-order-table


Hyperkid123:
cc @fionayhlin

src/smart-components/order/order-item.js
```diff
@@ -1,155 +1,121 @@
...
+        <Text className="pf-u-mb-0" component={TextVariants.small}>
+          <DateFormat date={item.created_at} variant="relative" />
+        </Text>
+      )
+    },
+    {
+      title: (
+        <Text className="pf-u-mb-0" component={TextVariants.small}>
+          {formatMessage(ordersMessages.lastUpdated)}
+          &nbsp;
```
rvsia, line 96: Can we put classname with a modifier to the next DateFormat?

src/smart-components/order/orders-list.js
```diff
@@ -233,66 +309,47 @@ const OrdersList = () => {
...
+                    {meta.noData
+                      ? formatMessage(ordersMessages.noOrdersTitle)
+                      : formatMessage(filteringMessages.noResults)}
+                  </Title>
+                  <EmptyStateBody>
+                    {meta.noData
+                      ? formatMessage(ordersMessages.noOrdersDescription)
+                      : formatMessage(filteringMessages.noResultsDescription)}
+                  </EmptyStateBody>
+
```
rvsia, line 340: empty line

src/smart-components/order/orders-list.js
```diff
@@ -233,66 +309,47 @@ const OrdersList = () => {
...
+                  <EmptyStateSecondaryActions>
+                    {!meta.noData && (
```
rvsia, line 342: what about to level-up this condition so `EmptyStateSecondaryActions` won't be empty?

src/smart-components/order/order-item.js
Hyperkid123: No need the last updated string should be removed!

Hyperkid123:
@Hyperkid123 all should be fine now.

Hyperkid123:
@rvsia I see your point. It's only doable by changing the columns definition. I have purposely put that inside a ref so it won't create the columns on each render. If only there was a feature to turn it off on the table level :no_mouth:. I will change it.

Hyperkid123:
I have forgot to add the sort fragment to the mocked URLs

PR #728, opened 2020-08-19
Edit edit option to add order process modal.

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1711

### Changes
- added edit variant to the add order process modal

Note: The order process and portfolio forms have exactly the same user flow and functionality. I will create a follow-up PR later which will merge these two components into a single component.

### DIFF
--- a/src/smart-components/order-process/add-order-process-modal.js
+++ b/src/smart-components/order-process/add-order-process-modal.js
@@ -1,5 +1,6 @@
 import React from 'react';
-import { useDispatch } from 'react-redux';
+import PropTypes from 'prop-types';
+import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import { useIntl } from 'react-intl';
 import { Modal } from '@patternfly/react-core';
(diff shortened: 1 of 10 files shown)


src/smart-components/order-process/add-order-process-modal.js
```diff
@@ -8,41 +9,77 @@ import { ORDER_PROCESSES_ROUTE } from '../../constants/routes';
...
     </Modal>
   );
 };
 
+AddOrderProcess.propTypes = {
+  edit: PropTypes.bool
+};
+
+AddOrderProcess.defaultProps = {
+  edit: false
```
rvsia, line 82: Do you really need this default prop?

rvsia (commented):
Just one code nitpick.

Otherwise, I found two issues:

**1**

image

I think the edit should be the first option.

**2**

When you enter the edit/add form, you lost the hash part of URL.

src/smart-components/order-process/add-order-process-modal.js
Hyperkid123: not really. It's just more explicit.

Hyperkid123:
@rvsia the order was changed and hash is now persistent. I would keep the default value.

rvsia (approved):
Tested the changes in UI, works great! :t-rex: Thank you!

PR #734, opened 2020-08-21
Add pre and post portfolio item selection to the add and edit order process modals

lgalis:
Add pre and post portfolio item selection to the add and edit order process modals

Implements https://projects.engineering.redhat.com/browse/SSP-1630

Screenshot from 2020-08-28 10-47-37
Screenshot from 2020-08-29 00-11-10

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "@patternfly/react-icons": "^4.7.1",
     "@patternfly/react-table": "^4.15.4",
     "@redhat-cloud-services/approval-client": "^1.0.64",
-    "@redhat-cloud-services/catalog-client": "^1.0.70",
+    "@redhat-cloud-services/catalog-client": "^1.0.71",
     "@redhat-cloud-services/frontend-components": "^2.3.14",
     "@redhat-cloud-services/frontend-components-notifications": "^2.1.2",
     "@redhat-cloud-services/frontend-components-utilities": "^2.1.3",
--- a/src/helpers/order-process/order-process-helper.ts
+++ b/src/helpers/order-process/order-process-helper.ts
@@ -28,11 +28,14 @@ export const loadProductOptions = (
 
   return getAxiosInstance()
     .get(
-      `${CATALOG_API_BASE}/portfolio-items?filter[name][contains]=${filterValue}&${initialLookupQuery ||
+      `${CATALOG_API_BASE}/portfolio_items?filter[name][contains]=${filterValue}&${initialLookupQuery ||
         ''}`
     )
     .then(({ data }) =>
-      data.map(({ id, name }) => ({ label: name, value: id }))
+      data.map((item: { name: string; id: string }) => ({
+        label: item.name,
+        value: item.id
+      }))
     );
 };
 
--- a/src/helpers/order-process/order-process-helper.ts
+++ b/src/helpers/order-process/order-process-helper.ts
@@ -50,11 +50,6 @@ export const fetchOrderProcess = (
     ApiCollectionResponse<OrderProcess>
   >;
 
-export const addOrderProcess = (
-  processData: OrderProcess
-): Promise<OrderProcess> =>
-  getOrderProcessApi().createOrderProcess(processData) as Promise<OrderProcess>;
-
 export const setOrderProcesses = (
   toTag: string[],
   toUntag: string[],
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -113,7 +113,7 @@
     "stylelint": "^13.5.0",
     "stylelint-config-recommended-scss": "^4.1.0",
     "ts-loader": "^8.0.2",
-    "typescript": "^4.0.2",
+    "typescript": "^3.9.7",
```
Hyperkid123, line 116: Was there any specific reason for this downgrade? TS v4 is not the lts version and we should use it.

src/helpers/order-process/order-process-helper.ts
```diff
@@ -18,6 +18,27 @@ export const listOrderProcesses = (
     `${CATALOG_API_BASE}/order_processes?filter[name][contains_i]=${filter}&limit=${limit}&offset=${offset}`
   );
 
+export const loadProductOptions = (
+  filterValue = '',
+  initialLookup: string[] = []
+) => {
```
Hyperkid123, line 24: This function is missing a return type.

src/helpers/order-process/order-process-helper.ts
```diff
@@ -73,8 +89,47 @@ export const removeOrderProcesses = (
...
+        )
+      : {};
+  const promiseA =
+    after_portfolio_item_id !== undefined
+      ? getOrderProcessApi().addOrderProcessAfterItem(
+          ((op as unknown) as OrderProcess).id as string,
+          { portfolio_item_id: after_portfolio_item_id }
+        )
+      : {};
+  return Promise.all([promiseA, promiseB]) as Promise<OrderProcess>;
```
Hyperkid123, line 134: This function surely does not return a promise with a single order process. Promise all always returns an array of responses in the same order as the in the array. This will require a custom type. IF the `addOrderProcessBeforeItem` returns a OrderProcess, then it should look something like this: `Promise<[OrderProcess, OrderProcess]>`

src/helpers/portfolio/portfolio-helper.ts
```diff
@@ -228,7 +228,7 @@ export const uploadPortfolioItemIcon = (
     return axiosInstance.patch(`${CATALOG_API_BASE}/icons/${iconId}`, data);
   }
 
-  data.append('portfolio_item_id', portfolioItemId);
+  data.append('portfolio_item', portfolioItemId);
```
Hyperkid123, line 231: Was this a bug you have found?

src/helpers/portfolio/portfolio-helper.ts
```diff
@@ -286,3 +286,8 @@ export const undeletePortfolio = (
   axiosInstance.post(`${CATALOG_API_BASE}/portfolios/${portfolioId}/undelete`, {
     restore_key: restoreKey
   });
+
+export const getPortfolioItemName = async (portfolioItemId: string) => {
```
Hyperkid123, line 290: Function missing return type.

src/utilities/constants.ts
```diff
@@ -26,3 +26,6 @@ export const APP_NAME = {
   PortfolioItem: 'catalog',
   ServiceInventory: 'topology'
 };
+
+export const BEFORE_TYPE = 'before';
+export const AFTER_TYPE = 'after';
```
Hyperkid123, line 31: These constant names do not match the `TYPE_BEFORE` and `TYPE_AFTER` imported in the schema.

Hyperkid123:
There is only an `after` select in the modal. Probably caused by the nonexisting constants.
before-after

lgalis:
@Hyperkid123  - this was still WIP - forgot to change the label - will address these issues and more on Tuesday.

src/helpers/order-process/order-process-helper.ts
```diff
@@ -18,6 +19,21 @@ export const listOrderProcesses = (
     `${CATALOG_API_BASE}/order_processes?filter[name][contains_i]=${filter}&limit=${limit}&offset=${offset}`
   );
 
+export const loadProductOptions = (
+  filterValue = ''
+): Promise<ApiCollectionResponse<PortfolioItem>> => {
```
Hyperkid123, line 24: This does not return `ApiCollectionResponse<PortfolioItem>`.

We can create a new common type form select options. Should look something like this:
```TS
export interface SelectOption {
  label: string;
  value?" any;
}

export type SelectOptions = SelectOption[]
```

src/helpers/order-process/order-process-helper.ts
```diff
@@ -73,8 +84,48 @@ export const removeOrderProcesses = (
...
+  });
+
+  return op as OrderProcess;
+};
+
+export const addOrderProcess = async ({
+  before_portfolio_item_id,
+  after_portfolio_item_id,
+  ...data
+}: Partial<OrderProcess>) => {
```
Hyperkid123, line 109: This return type should look like this:
```TS
tPromise<[OrderProcess, OrderProcess | undefined]
```

There might be only one promise, so the other OrderProcess› might be undefined.

src/helpers/order-process/order-process-helper.ts
```diff
@@ -73,8 +84,48 @@ export const removeOrderProcesses = (
...
+      : {};
+  const promiseA =
+    after_portfolio_item_id !== undefined
+      ? getOrderProcessApi().addOrderProcessAfterItem(
+          ((op as unknown) as OrderProcess).id as string,
+          { portfolio_item_id: after_portfolio_item_id }
+        )
+      : {};
+  await Promise.all([promiseA, promiseB]);
+  return op;
```
Hyperkid123, line 130: I think this should return the `Promise.all([promiseA, promiseB])` promise not the order process itself. The order process entity will be outdated. It will miss the processes ids and the updated_at flag will also be stale.

Hyperkid123 (commented):
The code works in the UI. There are just some missing TS types.

lgalis:
@Hyperkid123 - feedback addressed

Hyperkid123 (approved):

PR #735, opened 2020-08-24
Order page invalid access

Hyperkid123:
jira: https://projects.engineering.redhat.com/browse/SSP-1771

### Changes
- Fix UI crash when invalid order id is passed
  - if the API responds with 404 or 400, the user will be redirected to not found page
  - I have included 400 because the API treats noninteger ID as invalid and returns 400
- Allow sequential order data fetching if one or more order query params are undefined or null
  - allows simple access or order detail page from outside the application without the full context
- migrated handful of files to TS
  - constants
  - user-login
  - catalog-history


Hyperkid123:
@lgalis  is that after you click on an item or you manually add undefined to the URL?

lgalis:
@Hyperkid123   - just clicking an older order in the UI

Hyperkid123:
@lgalis ok Let me fix that as well.

Hyperkid123:
@lgalis alright undefined query params are no longer part of the order links.

PR #750, opened 2020-09-03
Migrate form schemas to TS

Hyperkid123:


src/forms/portfolio-share-form.schema.ts
```diff
@@ -1,7 +1,14 @@
+import Field from '@data-driven-forms/react-form-renderer/dist/cjs/field';
+import { SelectOptions } from '../types/common-types';
+import { Schema } from '@data-driven-forms/react-form-renderer';
```
rvsia, line 3: cjs import - I think you should be consistent, otherwise I hope that TS removes types imports from the final code. :smile: 

The same applies for other files in this PR.

Hyperkid123: VS code auto imports :+1:

rvsia (approved):

PR #751, opened 2020-09-03
Modal form template

Hyperkid123:
### Changes
- adds new modal form template
- removes additional scrollbars from modal forms
- uses correct PF modal layout (buttons do not overflow)
- was applied only on add/edit portfolio modal
  - Eventually, it will be applied to all modal forms but its a large task and I don't want to cram it into single PR

### Before
screenshot-ci cloud redhat com-2020 09 03-15_50_26

### After
screenshot-ci foo redhat com_1337-2020 09 03-15_30_55


src/forms/form-fields/form-template.tsx
```diff
@@ -0,0 +1,147 @@
...
+          </StyledForm>
+        </FocusTrap>
+      </Backdrop>
+    </div>,
+    document.body
+  );
+};
+
+export interface FormTemplateProps extends Pf4FormTemplateProps {
+  formContainer: 'default' | 'modal';
```
rvsia, line 130: Maybe we should change this to `isModal` to make the code a little bit simpler (if you don't expect more variants to come)

rvsia (approved):
Looks great! :taco: 

Just a small optimization suggestion and one possible code simplification.

src/forms/form-fields/form-template.tsx
```diff
@@ -0,0 +1,147 @@
...
+const FormTemplate: React.ComponentType<FormTemplateProps> = ({
+  formContainer = 'default',
+  modalProps = {},
+  ...props
+}) => {
+  if (formContainer === 'modal') {
+    return <ModalFormTemplate {...props} modalProps={modalProps} />;
+  }
+
+  return <PF4FormTemplate {...props} />;
```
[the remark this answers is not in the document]
Hyperkid123: We would have to add lazy loading and that will have ugly spinner flash if the loading flag is false.

src/forms/form-fields/form-template.tsx
Hyperkid123: We could do that. :+1:

rvsia (approved):

lgalis (approved):
Looks good, tested.

PR #752, opened 2020-09-03
Sprint 60 2nd merge stable to prod-stable

eclarizio:
I know this is kind of confusing, since https://github.com/RedHatInsights/catalog-ui/pull/745 looks like the same thing, but that one was basically merging up what was on stable from sprint 59 to prod-stable. This will make prod-stable reach parity with what was pushed to stable for sprint 60.

There will be *another* PR that eventually catches up stable (and thus later prod-stable) to what is being merged to prod-beta for Sprint 60 (all the code contained in https://github.com/RedHatInsights/catalog-ui/pull/747)


brahmanim (approved):
LGTM

PR #758, opened 2020-09-04
Migrate platform presentational components and order notification to TS.

Hyperkid123:


rvsia (approved):

rvsia:
@Hyperkid123  there are some typing errors

Hyperkid123:
@rvsia ah I forgot to add types to the placeholders file. It should be OK now.

Hyperkid123:
We have some snapshots mismatch. Taking a look.

Hyperkid123:
Ah these snapshot failures are caused by the change of the default props definition. If the `defaultProps` is used, then the prop is injected to the component from "above" and is seen in the virtual dom. If we assign the props when destructing the props the prop value only exists inside the function, therefore they are not visible in the virtual DOM.

lgalis (approved):

PR #770, opened 2020-09-16
Update workbox-webpack-plugin for security vulnerability fix

lgalis:
Fixes https://issues.redhat.com/browse/SSP-1838

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11787,12 +11787,6 @@
         "is-stream": "^1.0.1"
       }
     },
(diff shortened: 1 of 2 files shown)


Hyperkid123 (commented):
This issue one comes from the workbox plugin. Can you try and update the `workbox-webpack-plugin` dependency to `^5.1.4` instead of adding the self-signed directly?

Overriding it directly might not be necessary.

Hyperkid123 (approved):

PR #776, opened 2020-09-23
Limit the number of retries for approval_requests

lgalis:
When an order is stuck in Created state, the approval tab for an order will call get approval_requests in a loop.


eclarizio (commented):
I know this is WIP but a quick question: should there be any sort of action or at least logging for when we hit the max retries? I'm thinking it might seem odd that it just stops with no feedback to the user once the retry number has been hit.

PR #801, opened 2020-10-13
Update portfolio card description layout according to the new UX mock

lgalis:
Fixes https://issues.redhat.com/browse/SSP-1812

Screenshot from 2020-10-13 17-36-05


src/presentational-components/shared/card-common.tsx
```diff
@@ -2,9 +2,13 @@
 import React, { Fragment } from 'react';
 import styled from 'styled-components';
 import { AnyObject } from '../../types/common-types';
+import Truncate from 'react-truncate';
+import { Tooltip } from '@patternfly/react-core';
 
 const CardPropText = styled.div`
   overflow: hidden;
+  text-overflow: ellipsis;
+  lines: 3;
```
Hyperkid123, line 11: This CSS property works only in chrome and edge right now. I think `react-truncate` should work without it.

lgalis: @Hyperkid123 - you are correct, I meant to remove these lines.

Hyperkid123 (approved):

Hyperkid123:
@lgalis there is just a snapshot update.

PR #836, opened 2020-11-13
Fix for order processes sorting

lgalis:
Fixes https://issues.redhat.com/browse/SSP-1923

### DIFF
--- a/src/helpers/order-process/order-process-helper.ts
+++ b/src/helpers/order-process/order-process-helper.ts
@@ -8,19 +8,27 @@ import {
 import {
   ApiCollectionResponse,
   ApiMetadata,
-  SelectOptions
+  SelectOptions,
+  SortBy
 } from '../../types/common-types';
 import { AxiosResponse } from 'axios';
 const axiosInstance = getAxiosInstance();
 const orderProcessApi = getOrderProcessApi();
 
+const sortPropertiesMapper = (property: string): string => property;
+
 export const listOrderProcesses = (
   filter = '',
+  sortBy?: SortBy,
   { limit, offset }: ApiMetadata = defaultSettings
-): Promise<ApiCollectionResponse<OrderProcess>> =>
-  axiosInstance.get(
-    `${CATALOG_API_BASE}/order_processes?filter[name][contains_i]=${filter}&limit=${limit}&offset=${offset}`
+): Promise<ApiCollectionResponse<OrderProcess>> => {
+  const sortQuery = sortBy
+    ? `&sort_by=${sortPropertiesMapper(sortBy.property)}:${sortBy.direction}`
+    : '';
+  return axiosInstance.get(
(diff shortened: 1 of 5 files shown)


src/helpers/order-process/order-process-helper.ts
```diff
@@ -8,19 +8,27 @@ import {
...
   { limit, offset }: ApiMetadata = defaultSettings
-): Promise<ApiCollectionResponse<OrderProcess>> =>
-  axiosInstance.get(
-    `${CATALOG_API_BASE}/order_processes?filter[name][contains_i]=${filter}&limit=${limit}&offset=${offset}`
+): Promise<ApiCollectionResponse<OrderProcess>> => {
+  const sortQuery = sortBy
+    ? `&sort_by=${sortPropertiesMapper(sortBy.property)}:${sortBy.direction}`
+    : '';
+  return axiosInstance.get(
+    `${CATALOG_API_BASE}/order_processes?filter[name][contains_i]=${filter}&limit=${limit}&offset=${offset}${sortQuery}`
```
eclarizio, line 29: This might be a dumb request, but does it make sense instead to make line 29 contain the `&sort_by=` instead of line 26? So it would end up being `...&offset=${offset}&sort_by=${sortQuery}`?

All of the others in the parameter list seem to follow that pattern, so that's why it sticks out to me.

lgalis: I am only adding the sortBy parameter to the query if the user is sorting by a column. If I change it as sugessted, the url would be like:
https://ci.foo.redhat.com:1337/api/catalog/v1.2/order_processes?filter[name][contains_i]=&limit=50&offset=0&sort_by=

and it would result in an error.

eclarizio: Ah, that makes sense, cool 👍

src/smart-components/order-process/order-processes.js
```diff
@@ -185,9 +191,10 @@ const OrderProcesses = () => {
   const setSelectedOrderProcesses = (id) =>
     stateDispatch({ type: 'select', payload: id });
 
-  const updateOrderProcesses = (pagination) => {
+  const updateOrderProcesses = (pagination, sortBy) => {
+    console.log('Debug - updateOrderProcesses sortBy', pagination?.sortBy);
```
Hyperkid123, line 195: console log

src/smart-components/order/order-detail/order-provision.tsx
```diff
@@ -178,6 +178,7 @@ const OrderProvision: React.ComponentType = () => {
     formatMessage: FormatMessage,
     key: number
   ): RowType => {
+    console.log('Debug - item, progressMessages', item, progressMessages);
```
Hyperkid123, line 181: ^^

Hyperkid123 (approved):

PR #863, opened 2020-12-11
Only display the artifacts box when artifacts are present

lgalis:
https://issues.redhat.com/browse/SSP-1957

### DIFF
--- a/src/smart-components/order/order-detail/order-lifecycle.tsx
+++ b/src/smart-components/order/order-detail/order-lifecycle.tsx
@@ -55,7 +55,7 @@ const OrderLifecycle: React.ComponentType = () => {
           </CardBody>
         </Card>
       </StackItem>
-      {orderItem?.artifacts && (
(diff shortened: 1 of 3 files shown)


src/smart-components/order/order-detail/order-lifecycle.tsx
```diff
@@ -27,7 +27,11 @@ const OrderLifecycle: React.ComponentType = () => {
     ({ orderReducer: { orderDetail } }) => orderDetail || {}
   );
   const { order, orderItem } = orderDetailData;
-  if (order.state !== 'Completed' && order.state !== 'Ordered') {
+  if (
+    order.state !== 'Completed' &&
+    order.state !== 'Ordered' &&
+    order.state !== 'Failed'
```
Hyperkid123, line 33: I think we could use a shared function for this condition.

Hyperkid123 (approved):

PR #864, opened 2020-12-14
Update api path, separate order progress messages

lgalis:
https://issues.redhat.com/browse/SSP-2014

Update api path
Separate order progress messages
Add icon and color map entry for order 'Denied' status

Before:
Screenshot from 2020-12-14 13-55-10

After:
Screenshot from 2020-12-14 13-54-44

PR #867, opened 2020-12-14
[Prod-stable] Sprint 66 stable to prod-stable

eclarizio:
@brahmanim This includes a couple fixes from last week and today. Namely #864 for separating the progress messages, but also #863.

#866 updated stage to include the fixes, so once you've verified that stage is good, this will need to go out at the same time as the API (which can be found here: https://gitlab.cee.redhat.com/service/app-interface/-/merge_requests/13115)


brahmanim:
LGTM, waiting for API to be merged first

PR #929, opened 2021-02-17
Bump react core to 4.97.2 Updated snapshots

lgalis:


lgalis:
Version upgrade - merging

PR #953, opened 2021-03-17
Survey editor - Update initial value options  for the input field 

lgalis:
Fixes https://issues.redhat.com/browse/SSP-2167


Hyperkid123 (commented):
Rather than adding the value check to the useState, I would prefer adding default prop instead.

Hyperkid123 (approved):

PR #980, opened 2021-05-13
Fix for create order process with no 'after' portfolio item

lgalis:
https://issues.redhat.com/browse/SSP-2234

### DIFF
--- a/src/helpers/order-process/order-process-helper.ts
+++ b/src/helpers/order-process/order-process-helper.ts
@@ -121,13 +121,16 @@ export const updateOrderProcess = async (
   });
 
   let promiseB = {};
-  data.order_process_type === 'itsm'
-    ? (return_portfolio_item_id = undefined)
-    : (before_portfolio_item_id = after_portfolio_item_id = undefined);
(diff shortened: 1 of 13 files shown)


src/helpers/order-process/order-process-helper.ts
```diff
@@ -184,20 +187,26 @@ export const addOrderProcess = async ({
...
           ((op as unknown) as OrderProcess).id as string,
           { portfolio_item_id: after_portfolio_item_id }
         )
       : {};
   const promiseR =
+    return_portfolio_item_id &&
+    return_portfolio_item_id !== '' &&
     return_portfolio_item_id !== undefined
       ? getOrderProcessApi().addOrderProcessReturnItem(
           ((op as unknown) as OrderProcess).id as string,
```
eclarizio, line 212: Feels like all of these could be refactored into a single function, but I don't know javascript conventions very well, so I don't know how easy/difficult it is to do that.

eclarizio (approved):
Looks good to me, just one comment about a potential refactor. If it's quick then lets do it, but if it's more involved then maybe push out to a separate PR.

PR #986, opened 2021-05-20
[prod-beta] Merge main to prod-beta

lgalis:


eclarizio:
@lgalis Before we do this, can we fix all of the security issues that dependabot can't do automatically? One is high severity and the other two are medium, and it would be nice to get those knocked out

PR #1028, opened 2021-07-23
Update webpack and DDF

lgalis:


src/AppEntry.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Register simple service worker JS assets pre chache
+ * If registration fails, the app will continue to work withouth any issues.
+ * Service worker is generated by webpack and hould be ever avaiable only in production version of the app.
+ * You don't yant your JS bundles to be cached in dev env.
+ * You must test it locally in non HTTPS environment due to the insecure origin and self signed certificate issues cause by the PROXY.
+ * These cannot be overcome due to xss security policy inside service workers browser implementaion to avoid potential securitu issues.
+ */
+if (navigator.serviceWorker) {
```
Hyperkid123, line 23: You no longer need the service worker. It's not generated anymore and the chrome 2 performance improvement makes it not really worth it.

Hyperkid123 (commented):
I will do some local testing but those changes look very good.

The only thing I would also do is update react and react-dom to 17.0.x. That is the version that is chrome sharing to applications. There are no breaking changes between v16 and v17.

profiles/local-frontends.js
```diff
@@ -1,23 +1,41 @@
...
+  [`/${SECTION}/${APP_ID}/orders`]: {
+    host: `https://${localhost}:${FRONTEND_PORT}`
+  },
+  [`/beta/${SECTION}/${APP_ID}/orders`]: {
+    host: `https://${localhost}:${FRONTEND_PORT}`
+  },
   [`/apps/${APP_ID}`]: { host: `https://${localhost}:${FRONTEND_PORT}` },
-  [`/beta/apps/${APP_ID}`]: { host: `https://${localhost}:${FRONTEND_PORT}` }
+  [`/beta/apps/${APP_ID}`]: { host: `https://${localhost}:${FRONTEND_PORT}` },
+  [`/beta/config`]: { host: `http://localhost:8889` }
```
Hyperkid123, line 38: I believe this entry is here just for testing purposes?

Hyperkid123:
I've just tested it locally and it works perfectly!

profiles/local-frontends.js
lgalis: yes, testing only

lgalis:
@Hyperkid123 - I will update react to 17 in a separate PR - the tests will need to be updated, and this PR is already big

Hyperkid123:
@lgalis got it. Is there a CSC PR prepared for the switch?

Hyperkid123 (approved):

lgalis:
@Hyperkid123 - https://github.com/RedHatInsights/cloud-services-config/pull/821

Hyperkid123:
Ok, it should be ready in qa/ci-beta in a few minutes. We will have to coordinate for the rest of the environments. Let me know once you are ready.

PR #1046, opened 2021-08-19
Update survey editor breadcrumbs

lgalis:
Before:

Screenshot from 2021-08-19 10-51-32

After:
Screenshot from 2021-08-20 11-37-23


src/smart-components/portfolio/portfolio-item-detail/portfolio-item-detail.js
```diff
@@ -106,17 +106,21 @@ const PortfolioItemDetail = () => {
...
+            className="pf-u-mb-sm"
+            key={object}
+            variant="warning"
+            isInline
+            title={formatMessage(portfolioMessages.objectUnavaiable, {
+              object
+            })}
+          />
+        ))
+    : [];
```
eclarizio, line 123: It's possible I'm just really bad at reading javascript, but it feels very weird to me to see 10+ lines and a ternary operator. I'm not sure what the "javascript" way of cleaning this up would be, but maybe taking lines 110 through 122 into a separate function and simply calling that would make this easier to grok. It would probably all fit on line 109 then, as well.
```
const unavailable = portfolioItemData?.source ? <somethingWithAGoodName(portfolioItemData)> : []
```

src/smart-components/common/catalog-breadcrumbs.tsx
```diff
@@ -5,10 +5,18 @@ import ConditionalLink from '../../presentational-components/shared/conditional-
...
+  breadcrumbfragments?: BreadcrumbFragment[];
+}
+const CatalogBreadcrumbs: React.ComponentType<CatalogBreadcrumbsProps> = ({
+  breadcrumbfragments = []
+}) => {
+  const fragments = breadcrumbfragments?.length
+    ? breadcrumbfragments
+    : useSelector<CatalogRootState, BreadcrumbFragment[]>(
+        ({ breadcrumbsReducer: { fragments } }) => fragments
+      );
```
eclarizio, line 18: This ternary doesn't stand out as much to me, but I still feel like stretching a ternary over multiple lines is confusing.

eclarizio (commented):
Logic seems fine, I have a couple comments about the use of ternaries. I may just be out of touch with the javascript/typescript world, though, so if you want to push back and provide reasoning, I'll approve.

src/smart-components/common/catalog-breadcrumbs.tsx
lgalis: The TS code is verbose... however, I have updated both places where ternary op was used in this PR.

eclarizio (approved):

PR #1050, opened 2021-08-24
Left hand navigation for the standalone catalog

lgalis:
Add left hand standalone navigation, standalone header.
https://issues.redhat.com/browse/SSP-2301
https://issues.redhat.com/browse/SSP-2303

Screenshot from 2021-09-13 10-23-53
Screenshot from 2021-09-13 10-23-47
Screenshot from 2021-09-13 10-23-43
Screenshot from 2021-09-13 10-23-39
Screenshot from 2021-09-13 10-23-35
Screenshot from 2021-09-13 10-23-28


eclarizio (commented):
Does it make sense to have Order Processes at this point when we won't be implementing those in the on-prem version for quite a while? I think we might want to hide/remove that section for now

eclarizio (approved):

PR #1052, opened 2021-08-26
Add standalone webpack configuration

lgalis:
https://issues.redhat.com/browse/SSP-2300

Add a webpack config to build and run a standalone version of the catalog.

This is just the config PR - the standalone version first page is empty. The content will be added by follow-up PRs


eclarizio (approved):
This is probably better suited for a separate PR, but I think we should update the readme (or maybe the wiki?) with information on how to start the UI in a standalone fashion, which config files may need to be edited and how, and then which command(s) to actually run to get it working.

PR #1075, opened 2021-10-01
Fix for first sub-menu catalog navigation

lgalis:
Remove blocking for the first nav event, which was a workaround for chrome 1
Fixes https://issues.redhat.com/browse/SSP-2444

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -80,7 +80,7 @@ const App = () => {
        * Uses React history directly instead of browser history to avoid template realod.
        * only redirect after first application mount
        */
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -80,7 +80,7 @@ const App = () => {
        * Uses React history directly instead of browser history to avoid template realod.
        * only redirect after first application mount
        */
```
eclarizio, line 82: Does this explanation comment need updating since we aren't using `!ignoreRedirect` anymore?

eclarizio (approved):

PR #1150, opened 2022-01-28
Update Axios to v0.25.0

eclarizio:
https://issues.redhat.com/browse/SSP-2682

The 'follow-redirects' package is a dependency that axios relies upon, and thus they released v0.25.0 to update follow-redirects to a version with a fix for the above security flaw. https://github.com/axios/axios/releases/tag/v0.25.0


eclarizio:
@lgalis Looks like this update is going to break a lot of stuff unfortunately :(. Not a high priority, but will likely need your guidance to fix the build issues.

PR #1151, opened 2022-01-31
Revert "Bump @redhat-cloud-services/frontend-components from 3.5.1 to 3.6.1"

lgalis:
Reverts RedHatInsights/catalog-ui#1149


lgalis:
Reverting this as it causes issues with the filter button on the portfolio page ( PrimaryToolbar). 
Will update with a new version when this is fixed or addressed in the catalog UI .

PR #1153, opened 2022-02-02
Update workflow apis

lgalis:
Update workflow apis.
Updated routes.

Screenshot from 2022-02-02 08-49-37

### DIFF
--- a/src/standalone-nav.js
+++ b/src/standalone-nav.js
@@ -73,9 +73,6 @@ const App = (props) => {
       menuItem('Approval', {
         url: `${baseUrl}${Paths.approval}/index.html`
       }),
-      menuItem('Approval', {
-        url: `${baseUrl}${Paths.approval}`
-      }),
       menuItem(`Documentation`, {
         url:
           'https://access.redhat.com/documentation/en-us/red_hat_ansible_automation_platform/',
--- a/src/standalone-nav.js
+++ b/src/standalone-nav.js
@@ -50,16 +50,9 @@ const App = (props) => {
       type: 'item',
       name
     });
-    const baseUrl = 'https://catalog.k8s.local';
-    console.log(
-      'Debug catalog - window.location.pathname: ',
-      window.location.pathname
-    );
-    console.log('Debug catalog - window.location.href: ', window.location.href);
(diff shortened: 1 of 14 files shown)


src/standalone-nav.js
```diff
@@ -51,23 +50,25 @@ const App = (props) => {
       type: 'item',
       name
     });
-    const baseUrl = '';
+    const index = window.location.href.indexOf(window.location.pathname);
+    const baseUrl = window.location.href.substr(0, index);
+    console.log('Debug - calculated baseUrl', baseUrl);
```
eclarizio, line 55: Do we want to leave this in?

lgalis:
Removed the log.

eclarizio (changes requested):
For the actions and edit-approval-workflow components at least, I think there should be a corresponding test file. It would be a good idea to add one for the helper too

src/test/helpers/approval/approval-helper-s.test.js
```diff
@@ -0,0 +1,153 @@
...
+        object_id: '5'
+      }
+    ).then((data) => {
+      //expect(unlinkWorkflowsSpy).toHaveBeenNthCalledWith(1, [
+      //  'unlink-id-1',
+      //  {}
+      //]);
+      //expect(unlinkWorkflowsSpy).toHaveBeenNthCalledWith(2, 'unlink-id-2', {});
+      //expect(linkWorkflowSpy).toHaveBeenNthCalledWith(3, 'link-id-1', {});
+      //expect(linkWorkflowSpy).toHaveBeenNthCalledWith(4, 'link-id-2', {});
```
eclarizio, line 131: Are these just test expectations or does the `toHaveBeenNthCalledWith` not work as we want it to or something like that? I think if you want to leave these in as comments for bolstering this test later, that's probably fine, but we should have at least one expectation that gets asserted here, otherwise this test "passes" but only because there's nothing to fail on.

lgalis: @eclarizio - forgot to uncomment these afdter fixing the mock posts - test updated now

eclarizio (approved):

